PR #281, opened 2014-07-16
Fixed ssh-key auth

shermdog:
Added the ability to use IdentityFile from .ssh/config file
Also re-enabled default of underlying ncclient allow_agent and look_for_keys
By default ncclient will:
    1. Try passed in keyfile
    2. Look for keys loaded in agent
    3. Look for default keys in user home
    4. Try passed in user/pass combo

### DIFF
--- a/tests/functional/test_device_ssh.py
+++ b/tests/functional/test_device_ssh.py
@@ -1,0 +1,25 @@
+'''
+
+@author: rsherman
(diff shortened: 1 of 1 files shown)


vincentbernat:
Please, have a look at #648 as I wasn't happy with it was decided if the agent should be used.

PR #320, opened 2014-12-10
Add description field to op.ethport.EthPortView

krisamundson:
op.phyport.PhyPortView has a description field which allows one
to get() the interface descriptions. Since phyport does not
support aggregated ethernet interfaces(ae), this commit adds
description support so one can get these from ae interfaces.

Example:

```
from jnpr.junos.op.ethport import *
from jnpr.junos import Device
dev = Device(host='ex4550.example.com')
dev.open()
ports = EthPortTable(dev).get()
for port in ports:
    print port.description
```

### DIFF
--- a/lib/jnpr/junos/op/ethport.yml
+++ b/lib/jnpr/junos/op/ethport.yml
@@ -15,6 +15,7 @@ EthPortView:
   fields:
     oper: oper-status
     admin: admin-status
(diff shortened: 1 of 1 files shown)


krisamundson:
Heading towards the 6-month mark.  I will close this next week.

PR #327, opened 2014-12-21
Support 'proxycommand' from ssh_config in utils/scp

sergzin:
I would like to propose support for proxycommand from ssh config for copying files via jumphost.
os.path.expanduser resolve an issue if path to ssh_config specified as following '~/.ssh/config'


shermdog:
Jenkins: ok to test

PR #330, opened 2015-01-06
Added ConfigLoadError exception and nested rpc-error checking

shermdog:
Added ConfigLoadError exception and nested rpc-error checking


jeremyschulman:
nice!

PR #354, opened 2015-02-25
domain lookup in configuration first

vnitinv:
domain lookup in configuration followed up in \etc\resolv.conf then in \var\etc\resolv.conf

### DIFF
--- a/tests/unit/facts/test_domain.py
+++ b/tests/unit/facts/test_domain.py
@@ -3,7 +3,8 @@ __credits__ = "Jeremy Schulman"
 
 import unittest
 from nose.plugins.attrib import attr
-from mock import patch
+from mock import patch, MagicMock
+from lxml import etree
 
 from jnpr.junos.facts.domain import facts_domain
 from jnpr.junos import Device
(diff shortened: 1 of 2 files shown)


tests/unit/facts/test_domain.py
```diff
@@ -40,3 +43,20 @@ def test_resolv_conf_no_domain(self, mock_fs_cat):
         facts_domain(self.dev, self.facts)
         self.assertEqual(self.facts['domain'], None)
         self.assertEqual(self.facts['fqdn'], 'test')
+
+    @patch('jnpr.junos.facts.domain.FS.cat')
```
shermdog: This test should patch jnpr.junos.utils.fs.FS.cat with a return value appropriate to complete the function.  The test assertion should be the proper facts are set.

vnitinv: as per mock doc:
Imagine we have a project that we want to test with the following structure:
a.py
-> Defines SomeClass
b.py
-> from a import SomeClass
-> some_function instantiates SomeClass
Now we want to test some_function but we want to mock out SomeClass using patch. The problem
is that when we import module b, which we will have to do then it imports SomeClass from module
a. If we use patch to mock out a.SomeClass then it will have no effect on our test; module b already
has a reference to the real SomeClass and it looks like our patching had no effect.
The key is to patch out SomeClass where it is used (or where it is looked up ). In this case
some_function will actually look up SomeClass in module b, where we have imported it. The
patching should look like:
@patch(‘b.SomeClass’)

hence patch is done for domain.FS.cat as in domain FS has been imported.

assertion too has been carefully chosen as when we have mock_fs_cat.return_value = None then /etc/resolv.conf will return None and script to moves to check in /var/etc/resolv.conf

Do let me know if we can have it in some better way. will do the changes.

PR #357, opened 2015-03-02
Added SwRollbackError exception and logic to sw.rollback

### LINKED ISSUE #356: Utils.sw.rollback() not catching failure
The sw.rollback() function does not catch the following response from Junos
```<output>WARNING: Cannot rollback, /packages/junos.old is not valid</output>```

Also unit-test for rollback currently not functioning properly.

shermdog:
The sw.rollback() function previously did not return the correct information.
Added proper output for single and multi-re/vc systems and new SwRollbackError exception.
Resolves #356


shermdog:
Jenkins: retest this please

PR #360, opened 2015-03-20
Added op files to get transceiver light levels, temperature, etc

Jaredn:
Simple addition to lib/jnpr/junos/op/ to add support for gathering transmit/receive light level, temperature and voltage.  Can easily be expanded in fields_diag for thresholds if desired.


jnpr-community-netdev:

Issue #108, opened 2014-02-07
Enhance sw.install to supported mixed mode VC


sebastianw:
Mixed mode Virtual-Chassis require multiple packages as a set. Currently this isn't supported by the util.sw.install routine.

shermdog:
General note for implementation

* Will have to specify the packages for all types of VCF members
  * We will probably leave this up to the user to verify
* SCP those packages to master
* ```request system software add set [package1 package2 ...] reboot```

sebastianw:
@jeremyschulman Sorry for never getting back to you about this, somehow I forgot all about it. We have EX4500/EX4200 or EX4550/EX4200 mixed VCs. Sadly I can't remember the code we had but @shermdog has the right idea. :) In regards to verification I think `request system software add ...` will fail anyway if it has the wrong type(s) of software for the members. At least I hope it does, I don't think we ever tested this.

PR #365, opened 2015-04-06
Mixed VC installation feature

vnitinv:


vnitinv:
Will support #108 with this pull request

PR #366, opened 2015-04-07
Add l2circuit

fugitifduck:
add l2ckt schema to eznc


jnpr-community-netdev:

vnitinv:
@fugitifduck 
This is being merged to the main line of PyEZ.  In the future, all
additions for tables and view definitions will be in a separate repository that
can be rev'd separate from the main code.  Expect this change to happen by
early 2016.

Issue #388, opened 2015-06-17
<get-software-information> RPC output change causes exception in facts_software_version() on junos-install-* images


stacywsmith:
Prior to release 15.1, the output of the <get-software-information> RPC looks like:

user@r3> show version | display xml 
<rpc-reply xmlns:junos="http://xml.juniper.net/junos/14.2R3/junos">
    <software-information>
        <host-name>r3</host-name>
        <product-model>mx960</product-model>
        <product-name>mx960</product-name>
        <package-information>
            <name>junos-version</name>
            <comment>Junos: 14.2R3.8</comment>
        </package-information>
        <package-information>
            <name>junos</name>
            <comment>JUNOS Base OS boot [14.2R3.8]</comment>
        </package-information>
...

facts_software_version() was gathering the version info from:
xpath('software-information/package-information[normalize-space(name)="junos"]/comment')

Beginning in release 15.1, the output of the <get-software-information> RPC was modified to include a <junos-version> tag:

user@r2> show version | display xml 
<rpc-reply xmlns:junos="http://xml.juniper.net/junos/15.1R1/junos">
    <software-information>
        <host-name>robot</host-name>
        <product-model>mx104</product-model>
        <product-name>mx104</product-name>
        <junos-version>15.1R1.8</junos-version>    <<<<<<<<<<< new tag
        <package-information>
            <name>junos</name>
            <comment>JUNOS Base OS boot [15.1R1.8]</comment>   <<<<<<< old info still exists
        </package-information>

On traditional, jinstall* images, such as the output above, the <package-information> element for the "junos" package still exists, and there is no problem in facts_software_version().

However, in the new junos-install* images, there is no "junos"  <package-information> element:

user@r1> show version | display xml 
<rpc-reply xmlns:junos="http://xml.juniper.net/junos/15.1R1/junos">
    <software-information>
        <host-name>baku</host-name>
        <product-model>mx480</product-model>
        <product-name>mx480</product-name>
        <junos-version>15.1R1.8</junos-version>
        <package-information>
            <name>os-kernel</name>
            <comment>JUNOS OS Kernel 64-bit  [20150521.306001_builder_stable_10]</comment>
        </package-information>
        <package-information>
            <name>os-runtime</name>
            <comment>JUNOS OS runtime [20150521.306001_builder_stable_10]</comment>
        </package-information>
...

On these images, an IndexError exception is raised in facts_software_version():

Traceback (most recent call last):
[...]

PR #389, opened 2015-06-18
Address issue #388.

stacywsmith:
<get-software-information> RPC output changes in Junos releases >= 15.1.
First, try to determine the version from the <junos-version> tag, if
present. If not present, revert to the previous method of gathering the
version information from the "junos" package-information.


jnpr-community-netdev:

Issue #400, opened 2015-07-03
Device password setter doesn't work


stacywsmith:
The Device class includes a setter for the password property with the comment:
"Change the authentication password value.  This is handy in case the calling program needs to attempt different passwords."

However, this setter doesn't work:
```
>>> from jnpr.junos import Device
>>> r0 = Device(host='r0',password='incorrect')
>>> r0._password
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Device' object has no attribute '_password'
>>> r0._auth_password
'incorrect'
>>> r0.open()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/jnpr/junos/device.py", line 414, in open
    raise EzErrors.ConnectAuthError(self)
jnpr.junos.exception.ConnectAuthError: ConnectAuthError(r0)
>>> r0.password = 'user123'
>>> r0._password
'user123'
>>> r0._auth_password
'incorrect'
>>> r0.open()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/jnpr/junos/device.py", line 414, in open
    raise EzErrors.ConnectAuthError(self)
jnpr.junos.exception.ConnectAuthError: ConnectAuthError(r0)
```

It ends up setting the self._password attribute when it should have set the self._auth_password attribute.

PR #401, opened 2015-07-03
Address #400

stacywsmith:


jnpr-community-netdev:

PR #403, opened 2015-07-13
for JSON support

vnitinv:
Can be used like:

dev.rpc.get_route_information({'format': 'json'})
dev.rpc.get_configuration(dict(format='json'))

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -576,7 +576,7 @@ class Device(object):
         # From 14.2 onward, junos supports JSON, so now code can be written as
         # dev.rpc.get_route_engine_information({'format': 'json'})
 
-        if rpc_cmd_e.attrib.get('format') == 'json':
+        if rpc_cmd_e.attrib.get('format') in ['json', 'JSON']:
             return json.loads(rpc_rsp_e.text)
 
         # This section is here for the possible use of something other than ncclient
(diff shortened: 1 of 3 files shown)


shermdog:
The key for this should be something like junos-json or some other descriptor that indicates this is the native Junos version of json vs. pure json.

Also this PR has not associated unit tests.

vnitinv:
I thought to keep key in sync with what we use with other formats. Ex:
dev.rpc.get_route_engine_information({'format': 'text'})

For me with dev.rpc.get_route_engine_information({'format': 'json'}) it should mean native junos json only.
Defining our own new key will add overhead. Above syntax is what goes into xml rpc.

Missed to add UT for this change. Will do so. Thanks @shermdog

vnitinv:
Jenkins: retest this please.

b1naryth1ef:
Whats the ETA of this being mainline?

vnitinv:
@b1naryth1ef I am expecting this to be released in a weeks time.

b1naryth1ef:
It also looks like the version pinning of this is incorrect. We have a QFX VCF on JUNOS 14.1X53-D27.3  which supports JSON RPC (I've tested both on the command line and by patching this PR), however this PR will not allow us to use that functionality. Can you take a look and confirm what versions actually support JSON RPC / whether this PR will support 14.1?

vnitinv:
@b1naryth1ef Let me check this with corresponding team.

vnitinv:
@b1naryth1ef We will support it from 14.2 onwards devices as for 14.1 its not consistent across platforms. 
As you are aware we do check for version on facts and facts is mutable, hence even for 14.1 json can be called by setting facts before rpc call. So you will have a hack to achieve the same.

PR #418, opened 2015-07-30
Prefer explicitly defined ssh_private_key_file to sshconfig

thomseddon:
The previous implementation had two issues:

1) Even if ssh_private_key_file is explicitly passed to the Device constructor, this would be overridden if identityfile was set in sshconfig, contrary to the many in which "user" was handled whereby the value passed to the constructor is preferred

2) If a matching statement was found in sshconfig then even in the case where identityfile had not been set, the value explicitly passed to the constructor would be ignored

This change makes the way in which ssh_private_key_file is handled the same as the way "user" is handled by always preferring a value explicitly passed

I have the same 6 tests failing before and after this patch (mostly related to being unable to reach the *.englab.juniper.net hosts)

I think the test for ssh_private_key_file would not have actually tested the correct behaviour before as if valid credentials were found in sshconfig then these would always have been used to open the connection

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -267,7 +267,7 @@ class Device(object):
             self._hostname = found.get('hostname', self._hostname)
             self._port = found.get('port', self._port)
             self._conf_auth_user = found.get('user')
-            self._ssh_private_key_file = found.get('identityfile')
+            self._conf_ssh_private_key_file = found.get('identityfile')
             return sshconf_path
 
     def __init__(self, *vargs, **kvargs):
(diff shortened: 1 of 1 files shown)


jnpr-community-netdev:

stacywsmith:
@thomseddon 
I like it. I had noticed the same issue and hadn't gotten around to reporting or coming up with a diff. Thanks for reporting.

thomseddon:
Ah, tests were failing - fixed now :)

vnitinv:
Thanks @thomseddon It would be good if we add unit test for this changes to make it error proof with changes coming in future. Would you like to contribute for the same.

PR #433, opened 2015-09-10
implement context manager for Device

GIC-de:
Implementation of context manager for Device (e.g. `with Device(...) as dev:`). 

```
>>> from jnpr.junos import Device
>>>
>>> with Device(host="1.2.3.4", user="user", port=22) as dev:
...     print "conntected to %s" % dev._facts["hostname"]
...     chassis = dev.rpc.get_chassis_inventory()
...     print "device: %s " % chassis.findtext(".//chassis/description")
...
conntected to BNGJMX3
device: MX960
```
The `__enter__` method will open the connection of the device and `__exit__` close the connection except a `ConnectError`exception was raised. The assumption is that the connection was already closed in this case and an additional close raises another exception.

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -799,3 +799,15 @@ class Device(object):
             probe_ok = False
 
         return probe_ok
+
+    # -----------------------------------------------------------------------
+    # Context Manager
+    # -----------------------------------------------------------------------
+
+    def __enter__(self):
+        self.open()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if not isinstance(exc_val, EzErrors.ConnectError):
+            self.close()
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -809,5 +809,5 @@ class Device(object):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        if not isinstance(exc_val, EzErrors.ConnectError):
+        if self.connected and not isinstance(exc_val, EzErrors.ConnectError):
             self.close()
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -809,5 +809,6 @@ class Device(object):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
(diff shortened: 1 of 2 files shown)


jnpr-community-netdev:

vnitinv:
Hi @GIC-de Nice feature to add. But can you please add some of the checks.
For example if the device the gets disconnected in the code flow then __exit__ close would break.
So please add self._conn.connected value check before calling close. 

```python
def __exit__(self, exc_type, exc_val, exc_tb):
    if self._conn.connected and isinstance(exc_val, EzErrors.ConnectError):
        self.close()
```

example where it can break
```python
with Device(host='xxxx, user='xxxx', password='xxxx') as dev:
    sw = SW(dev)
    ok = sw.install(pkg)
    if ok:
          sw.reboot()
          time.sleep(30)
```
Now once software is installed then we call reboot. This will disconnect the connection. Hence if the __exit__ close will break. So before calling close do check the connection as mentioned.

Please do add test cases (if you can) for the same. This will help to make this code changes robust.

GIC-de:
Hi @vnitinv, to prevent closing an already closed connection I checked if `__exit__` was called by an exception of `EzErrors.ConnectError`. Adding an check of `self._conn.connected` makes sense in a few cases like mentioned in your comment. Also if developers close the connection within the context by mistake. I will also add some tests cases.

GIC-de:
Hi @vnitinv, the `__exit__ ` method checks for `self.connected` instead of `self._conn.connected`. Actually `self._conn.connected` would be better as it covers also the case of lost SSH connections without raised `EzErrors.ConnectError` but I would prefer to implement a getter `@property` for `connection` returning `self._conn.connected` (like example below ...). 

```
@property
def connection(self):
    try:
        return self._conn.connected
    except AttributeError:
        return False
```

vnitinv:
hi @GIC-de In my comment above I too mentioned to use self._conn.connected only

do use self._conn.connected. I would't suggest adding connection property as Device already have connected attribute. Would like to change this only but would be risky as lot of user are already using it.
So you go ahead and use self._conn.connected only.

GIC-de:
Hi @vnitinv, is the current implementation okay now?

vnitinv:
@GIC-de Looks good to me. Will do some sanity testing before merging it.
Thanks for your pull request.

Issue #329, opened 2015-01-02
Enforce facts/chassis.py to support new hardware


titom73:
Hello,

In some cases, serial-number is not reported to chassis/serial-number neither in midplane. So when we get facts from device, exception is raised:

```
>>> from jnpr.junos import Device
>>> dev = Device(host='73.1.2.9', user='svc-netconf', password='nomios101')
>>> dev.open()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\tgrimonet\AppData\Roaming\Python\Python27\site-packages\jnpr\junos\device.py", line 439, in open
    self.facts_refresh()
  File "C:\Users\tgrimonet\AppData\Roaming\Python\Python27\site-packages\jnpr\junos\device.py", line 660, in facts_refresh
    gather(self, self._facts)
  File "C:\Users\tgrimonet\AppData\Roaming\Python\Python27\site-packages\jnpr\junos\facts\chassis.py", line 56, in facts_chassis
    'chassis-module[name="Backplane"]/serial-number')[0].text
IndexError: list index out of range
```
Here is complete ```show chassis hardware | display xml``` from impacted device and then from Firefly SRX

```
<rpc-reply xmlns:junos="http://xml.juniper.net/junos/14.1R1/junos">
    <chassis-inventory xmlns="http://xml.juniper.net/junos/14.1R1/junos-chassis">
        <chassis junos:style="inventory">
            <name>Chassis</name>
            <description>VMX</description>
            [...]
            <chassis-module>
                <name>FPC 0</name>
                <description>10x1G FPC</description>
                <chassis-sub-module>
                    <name>CPU</name>
                    <version>Rev. 1.0</version>
                    <part-number>RIOT</part-number>
                    <serial-number>123XYZ987</serial-number>
                </chassis-sub-module>
        [...]
        </chassis>
    </chassis-inventory>
    <cli>
        <banner></banner>
    </cli>
</rpc-reply>
```
SRX hardware:
```
<rpc-reply xmlns:junos="http://xml.juniper.net/junos/12.1X47/junos">
    <chassis-inventory xmlns="http://xml.juniper.net/junos/12.1X47/junos-chassis">
        <chassis junos:style="inventory">
            <name>Chassis</name>
            <serial-number>6cf0dcb880e9</serial-number>
            <description>FIREFLY-PERIMETER</description>
            [...]
        </chassis>
    </chassis-inventory>
    <cli>
        <banner></banner>
[...]

PR #434, opened 2015-09-25
Handle missing data for serial-number fact on vMX

cmek:
vMX doesn't have the serial-number tag under the chassis one. Additionally, there is no chassis module called Backplane so the device.open() fails with IndexError exception in:
 facts['serialnumber'] = x_ch.xpath(
                  'chassis-module[name="Backplane"]/serial-number')[0].text

### DIFF
--- a/lib/jnpr/junos/facts/chassis.py
+++ b/lib/jnpr/junos/facts/chassis.py
@@ -43,10 +43,20 @@ def facts_chassis(junos, facts):
 
     facts['model'] = x_ch.findtext('description')
 
-    try:
-        facts['serialnumber'] = x_ch.find('serial-number').text
-    except:
+    facts['serialnumber'] = x_ch.get('serial-number', '')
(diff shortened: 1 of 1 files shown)


jnpr-community-netdev:

vnitinv:
@cmek Do have a look to #329 It seems you are trying to solve the same issue.

cmek:
Looks like the right fix for this depends on what the definition of the serialnumber fact is - which component's serial number should be treated as the device serial number in case of different hardware architectures?

I'm happy to modify the patch to implement the correct logic depending on the serialnumber's definition but I'd need some guidelines on what the definition is.

Alternatively, the proposed patch could be modified to simply add try/except around facts['serialnumber'] = x_ch.xpath(
'chassis-module[name="Backplane"]/serial-number')[0].text
to avoid raising the exception and without changing the current logic of what the fact is and breaking backward compatibility. It would also fix #329.

PR #441, opened 2015-10-10
Added Table/View for IDP attack table (for SRX devices)

pklimai:
Created a Table/View for IDP attack table. Should be useful for gathering IDP attack statistics from the SRX.


jnpr-community-netdev:

vnitinv:
@pklimai 
This is being merged to the main line of PyEZ.  In the future, all
additions for tables and view definitions will be in a separate repository that
can be rev'd separate from the main code.  Expect this change to happen by
early 2016.

PR #444, opened 2015-10-14
Fix dev.cu.load() to throw RpcTimeoutError when encountered

### LINKED ISSUE #443: Config load not throwing RpcTimeoutError
When encountering a timeout while doing a dev.cu.load(), an RpcTimeoutError should be thrown,
instead it currently attempts to throw a ConfigLoadError and crashes out.

JamesNickerson:
fixes #443


jnpr-community-netdev:

Issue #432, opened 2015-09-07
Inconsistent start_shell.run() return value 


akbhat:
The tuple returned by run() in start_shell.py seems to vary making it difficult to parse the results. For example three consecutive runs of the shell command below to get the pid returned three slightly different tuples.

["ps -auxw | grep pfe | grep /usr/sbin | awk '{print $2}'", '\r\r\n', '1393\r\n1514\r\n', '% ']
["ps -auxw | grep pfe | grep /usr/sbin | awk '{print $2}'", '\r\r\n', '1393\r\n', '1514\r\n', '% ']
["ps -auxw | grep pfe | grep /usr/sbin | awk '{print $2}'", '\r\r\n', '1393\r\n1514\r\n% ']

stacywsmith:
A quick look at the code shows that run() doesn't actually comply with it's documentation. Specifically, the return value is not as documented. 

The documentation says:

    def run(self, command, this=_SHELL_PROMPT):
        """
        Run a shell command and wait for the response.  The return is a
        tuple. The first item is True/False if exit-code is 0.  The second
        item is the output of the command.
        :param str command: the shell command to execute
        :param str this: the exected shell-prompt to wait for
        :returns: result of the executed shell command (str)
        .. note:: as a *side-effect* this method will set the ``self.last_ok``
                  property.  This property is set to ``True`` if ``$?`` is
                  "0"; indicating the last shell command was successful.
        """

In reality, run() does set self.last_ok, but it does not return True/False as the first item in the tuple. It also doesn't return a string. It returns a tuple of strings. The strings are randomly broken up, as @akbhat reported, as they are read from the connection by the wait_for() method. The concatenation of this tuple is the full string which should be returned to the user.

I haven't tested, but I believe this method could be fixed to comply with the documentation by:

Change line 113 of lib/jnpr/junos/utils/start_shell.py from:
got = self.wait_for(this)

to:
got = ''.join(self.wait_for(this))

and change line 120 from:
return got
to
return (self.last_ok,got)

Since this would change the behavior of this method, we would need to double check for any code that currently uses the run() method....

vnitinv:
Hi @akbhat How about you sending the pull request with above mentioned changes. We can test the same and merge if all goes well.

EusaHues:
+1 on the inconsistent return tuples being "unhelpful". That part at least needs fixed.
Please don't just "fix" by updating the function document to match current (unhelpful/broken) behavior.

[...]

PR #445, opened 2015-10-14
Change the return value of start_shell.run() to comply with docs

stacywsmith:
Addresses #432

Change the return value of start_shell.run() to comply with the documentation:

    "The return is a tuple. The first item is True/False if
     exit-code is 0.  The second item is the output of the command."

Change a couple of places that were expecting the old return value.

### DIFF
--- a/lib/jnpr/junos/utils/fs.py
+++ b/lib/jnpr/junos/utils/fs.py
@@ -394,8 +394,7 @@ class FS(Util):
     def _ssh_exec(self, command):
         with StartShell(self._dev) as sh:
             got = sh.run(command)
-            ok = sh.last_ok
-        return (ok, got)
+        return got
 
     def rmdir(self, path):
         """
--- a/lib/jnpr/junos/utils/start_shell.py
+++ b/lib/jnpr/junos/utils/start_shell.py
@@ -17,8 +17,7 @@ class StartShell(object):
         def _ssh_exec(self, command):
             with StartShell(self._dev) as sh:
                 got = sh.run(command)
-                ok = sh.last_ok
-            return (ok, got)
+            return got
 
     """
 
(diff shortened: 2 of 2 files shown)


jnpr-community-netdev:

vnitinv:
@stacywsmith Lot os user already using this feature by handling output as list. Should we change documentation rather the return in sync to doc. I do understand inconsistency in outputs #432 . But we should fix this somehow rather than changing return type.

stacywsmith:
I submitted the pull request because you asked for it.

> Hi @akbhat How about you sending the pull request with above mentioned changes.
> We can test the same and merge if all goes well.

Given the current behavior I question your assertion that "lot of user already using this feature", but I do understand your concern about backwards compatibility.

vnitinv:
My bad @stacywsmith . Let me keep it open. Will test the same. Will check if can have any solution to inconsistency or not.  Inconsistency in output of string can lead to issues with current user's scripts as well. If the changed behavior of making return value as string from list makes more sense, We can ask existing user of this functionality to change their script for obvious reason.

jnpr-community-netdev:

stacywsmith:
Hi @vnitinv,

Here's what the documentation currently says:

        """
        Run a shell command and wait for the response.  The return is a
        tuple. The first item is True/False if exit-code is 0.  The second
        item is the output of the command.
        :param str command: the shell command to execute
        :param str this: the exected shell-prompt to wait for
        :returns: result of the executed shell command (str)
        .. note:: as a *side-effect* this method will set the ``self.last_ok``
                  property.  This property is set to ``True`` if ``$?`` is
                  "0"; indicating the last shell command was successful.
        """

However, the current behavior is that this method does not return the True/False value in the first item of the tuple. It only returns a tuple of strings which are randomly split substrings of the full output.

One potential solution might be to always return a tuple with a single item which is the full (concatenated) output string. This would be compatible with the current behavior and would avoid users ever having to do the concatenation of strings themselves. We would also want to update the documentation to correctly reflect the new behavior.

Thoughts?

EusaHues:
I'm using this change locally to get sane behavior from start_shell.

Issue #397, opened 2015-06-29
Wrap facts gathering in try block


shermdog:
The default gathering of facts occasionally fails due to the fragile nature of fact gathering.

Wrap the facts.refresh() in a try block and print a warning to the user on failure (include the full stack for debug).

PR #459, opened 2015-12-04
facts gathering should not raise and break scripts

vnitinv:
handle facts gathering and raise only warning in case of exception. So that script can continue with activities say if its not dependent on facts. For ex dev.rpc.rpc_calls() are not dependent on facts.


vnitinv:
Resolves #397

PR #462, opened 2015-12-10
Catch case where RPCError object is a list, not XML

shelson:
When committing configuration on a SRX, there are times where the operation can throw multiple RPCErrors. ncclient returns these as a list (by design) which trips up device.py when checking these errors, causing an unhandled exception as opposed to returning the error message to the user.

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -564,7 +564,12 @@ class Device(object):
         except NcErrors.TransportError:
             raise EzErrors.ConnectClosedError(self)
         except RPCError as err:
-            rsp = JXML.remove_namespaces(err.xml)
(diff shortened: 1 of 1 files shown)


shelson:
Gist showing before/after output: https://gist.github.com/shelson/443f70ab7eff9dd3461e

vnitinv:
@shelson We already have fixed this in github. Did you installed PyEZ from git or pypi?

Give a try by installing PyEZ from github (using below command)
pip install git+https://github.com/Juniper/py-junos-eznc.git

Do let us know if the issue is seen with latest git code too

shelson:
Cool! Not sure what happened there but a clean virtualenv does work with the latest git code, something must have gotten confused in my testbed.

Thanks for fixing it!

PR #475, opened 2016-02-04
'routing_engine' argument added to Device class which will allow to connec…

gaddevijay:
'routing_engine' argument added to Device class which will allow to connect to desired RE
Values to this argument can be master/backup/re0/re1/any

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -430,7 +430,7 @@ class Device(object):
             with StartShell(self) as sh:
                 got = sh.run("sysctl hw.re.slotid")
                 if not sh.last_ok:
-                    return None
+                    raise Exception("Could not determine routing engine slot id")
                 # get slot id from the result
                 got = '\n'.join(got)
                 mat = re.search('hw.re.slotid:\s*(\d+)', got, re.IGNORECASE)
@@ -537,6 +537,7 @@ class Device(object):
             if gather_facts is True and self._RE is not None:
                 self._connected_re = _get_connected_slot()
             if self._RE:
+                self._connected_re = _get_connected_slot()
                 _connect_to_re()
 
         except NcErrors.AuthenticationError as err:
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -10,6 +10,7 @@ import socket
 import datetime
 import time
 import json
+import re
 
 # 3rd-party packages
 from lxml import etree
@@ -27,6 +28,7 @@ from jnpr.junos.cfg import Resource
 from jnpr.junos.facts import *
 from jnpr.junos import jxml as JXML
 from jnpr.junos.decorators import timeoutDecorator, normalizeDecorator
+from jnpr.junos.utils.start_shell import StartShell
 
 _MODULEPATH = os.path.dirname(__file__)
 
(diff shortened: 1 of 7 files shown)


lib/jnpr/junos/facts/chassis.py
```diff
@@ -41,6 +46,16 @@ def facts_chassis(junos, facts):
         facts['2RE'] = False
         x_ch = rsp.find('chassis')
 
+    if x_ch is None:  # This is when backup RE is being connected
+        try:
+            rsp = junos.rpc.get_software_information()
+            if rsp.tag == 'error':
+                raise RuntimeError()
```
vnitinv: Add message in runtime error. Or define an appropriate class exception for such cases.

lib/jnpr/junos/device.py
```diff
@@ -441,6 +528,18 @@ def open(self, *vargs, **kvargs):
...
 
+            # If routing_engine(_RE) is defined then check if you are
+            # connected to correct RE, else get name of other re and
+            # connect to it.
+            self.connected = True
+
+            gather_facts = kvargs.get('gather_facts', self._gather_facts)
+            if gather_facts is True and self._RE is not None:
+                self._connected_re = _get_connected_slot()
+            if self._RE:
```
vnitinv: check the logic

lib/jnpr/junos/device.py
```diff
@@ -291,6 +293,12 @@ def __init__(self, *vargs, **kvargs):
         :param int port:
             *OPTIONAL* NETCONF port (defaults to 830)
 
+        :param str routing_engine:
+            *OPTIONAL* If not provided default behavior will be raise exception
```
vnitinv: change the doc string a bit as discussed

lib/jnpr/junos/device.py
```diff
@@ -412,6 +422,99 @@ def open(self, *vargs, **kvargs):
...
+        #  PRIVATE Functions
+        #
+
+        def _get_connected_slot():
+            """return connected slot. In case of error return None"""
+
+            rsp = self.rpc.get_chassis_inventory()
+            if rsp.tag == 'error':
+                raise RuntimeError('Could not determine RE slot id')
+            if rsp.tag == 'output':
```
stacywsmith: I can confirm this is an incorrect assumption for an MX VC. On an MXVC, where each member has dual REs, all 4 routing engines successfully execute the get-chassis-inventory RPC.

lib/jnpr/junos/device.py
```diff
@@ -412,6 +422,99 @@ def open(self, *vargs, **kvargs):
...
+                self._reRole = "backup"
+            else:
+                self._reRole = "master"
+
+            rsp = self.rpc.get_route_engine_information()
+            if rsp.tag == 'error':
+                raise RuntimeError('Could not determine RE slot id')
+            re_list = rsp.xpath('.//route-engine')
+            if len(re_list) == 1:
+                return "RE0"
```
stacywsmith: This is an incorrect assumption. While uncommon, a chassis can have a single RE installed in slot 1. If there's only one RE, you should still look at the <slot> tag to determine the correct slot.

lib/jnpr/junos/device.py
```diff
@@ -412,6 +422,99 @@ def open(self, *vargs, **kvargs):
...
+            for re in re_list:
```
stacywsmith: This loop doesn't correctly handle any multi-chassis system.

lib/jnpr/junos/device.py
```diff
@@ -412,6 +422,99 @@ def open(self, *vargs, **kvargs):
...
+
+            # Determine other re
+            other_re = None
+            if self._connected_re.upper() == 'RE0':
+                other_re = 're1'
+            else:
+                other_re = 're0'
+
+            # get the name of other re
+            command = etree.XML('<configuration><groups><name>%s</name>'
```
stacywsmith: This assumes the RE has a unique hostname configured, and the host executing the PyEZ script can resolve this unqualified hostname to a unique IP address, has connectivity to that IP address, etc. There are a LOT of assumptions here which I don't think are safe to make.

jnpr-community-netdev:

vnitinv:

vnitinv:
Thanks for the pull request. will do it properly in another pull request in future.

PR #476, opened 2016-02-09
FTP utility

gaddevijay:
FTP utility can be used to transfer files to and from device
This is required when Telnet and console connections are implemented and user wants to use FTP to transfer files.


jnpr-community-netdev:

PR #482, opened 2016-03-03
Python 3.x support

vnitinv:
With this pull request we can support python 3.x versions too.


pklimai:
Hi Nitin

Does it mean PyEz now supports Python 3? Will any doc or examples be available?

- Peter

vnitinv:
@pklimai Yes, we will support it from next major release.

PR #486, opened 2016-03-23
Improve the auto-detection of "text" config format.

joshainglis:
Improve the auto-detection of "text" config format.

Regex contained an error in which the `{` was not escaped which
caused an exception. Regex also failed to catch tokens split with `-`.


joshainglis:
Moved pull to https://github.com/Juniper/py-junos-eznc/pull/487

PR #488, opened 2016-03-29
Set table/view support.

ganeshnalawade:
Add support for set table/view which enables configuration in structured  manner.


vnitinv:
<xyz key=value>
Can you check how user can pass values to tag attributes?

PR #507, opened 2016-05-06
I431

spidercensus:


spidercensus:
Added a timeout of 60 seconds and created a new Exception class to handle the timeout.

PR #515, opened 2016-05-25
Added cryptography package for pip install

ntwrkguru:
@vnitinv please review and pull if appropriate

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ jinja2>=2.7.1
 PyYAML>=3.10
 netaddr
 six
+cryptography


vnitinv:
@ntwrkguru Neither PyEZ or its dependent modules requirement is cryptography (if I am not wrong). The issue 
```AttributeError: 'EntryPoint' object has no attribute 'resolve'```
some of us faced is due to an old version of setuptools (which uses cryptography)
https://github.com/pyca/cryptography/issues/2853

Said that, still do you think we should merge this? We can keep it open if other users faces same issue. Then we can take a call.

**Comment by ntwrkguru** [comment]
I'll try a fresh install in vagrant tonight. I remember this causing me some grief in 14.04 with setuptools 3.3. Looks like ncclient installs cryptography, so it shouldn't be needed. I'll comment later.

**Comment by stacywsmith** [comment]
I don't believe we need to list cryptography as a direct dependency in PyEZ.

cryptography is already indirectly listed as a dependency by:
PyEZ requires ncclient>=0.4.6
https://github.com/Juniper/py-junos-eznc/blob/master/requirements.txt#L2

ncclient requires paramiko>=1.7.7.1
https://github.com/ncclient/ncclient/blob/master/requirements.txt#L2

paramiko requires cryptography>=1.1
https://github.com/paramiko/paramiko/blob/master/setup.py#L79

As the link Nitin provided specifies, the issue is between cryptography and setuptools. cryptography >= 1.0 requires setuptools >=11.3
https://github.com/pyca/cryptography/commit/30e199e0fb220a17b3089508a59d29a2924f10e4

The bottom line is that I believe all of the dependencies are already correctly specified.

However, I did experience this problem on a freshly installed and updated:
user@h0:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 8.4 (jessie)
Release:	8.4
Codename:	jessie

The problem appears to be from following the installation instructions at:
http://www.juniper.net/techpubs/en_US/junos-pyez1.2/topics/task/installation/junos-pyez-server-installing.html

Specifically, I installed the python-pip package. This installs older versions of setuptools and cryptography. Because they are installed from an OS package, they also fail to get properly upgraded by the pip install of PyEZ.

I worked around this by installing pip from source rather than the python-pip package:
wget https://bootstrap.pypa.io/get-pip.py -O - | sudo python

**Comment by sdndude** [comment]
@stacysmith 
So to fix this, you either have to specify the right cryptography version or get around it by installing pip from source so that it upgrades cryptography when it is needed.  This is one of those things where we can control the former and the latter is tribal knowledge.  I vote for adding cryptography to the requirements instead of hoping everyone figures out the workaround with pip.  
Probably not the best solution, but it is the most scalable.

**Comment by vnitinv** [comment]
@sdndude @stacywsmith Even if we add cryptography in PyEZ (on top of it already being in cryptography requirements) requirements and if already an older version is present it won't install it. If we do -U (upgrade) then even paramiko should ideally take care of it.

Not sure what would be the best solution.
Should we raise an issue with Paramiko?

**Comment by stacywsmith** [comment]
I'll try to test further and verify, but I believe @vnitinv is correct. Adding a cryptography requirement in PyEZ will not fix this problem.

The cryptography requirement is already correctly specified in paramiko. The required version is not met (at least in my case), but the upgrade did not occur because the cryptography and setuptools libraries were installed by the OS packaging system and not pip.

Specifying cryptography as a requirement in PyEZ would not change this behavior.

IMO the issue is with pip. It seems to report a successful installation even though the upgrade to the required version of the prerequisite library failed.

**Comment by vnitinv** [comment]
This issue requires below command to be run on Linux boxes.

sudo yum install gcc libffi-devel python-devel openssl-devel
or
sudo apt-get install libffi-devel, openssl-devel

Will update required doc accordingly 
https://github.com/Juniper/py-junos-eznc/blob/master/INSTALL-UBUNTU-DEBIAN.md
https://github.com/Juniper/py-junos-eznc/blob/master/INSTALL-FEDORA.md

**Comment by ntwrkguru** [comment]
Nitin,

So, I've been messing around with this and the revised instructions do not provide a clean install on 14.04. Here's the recipe that I've been able to use reliably:

```
sudo apt-get update && sudo apt-get install -y \
    build-essential \
    python-setuptools \
    python-dev \
    libxml2-dev \
    libxslt-dev \
    libssl-dev \
    libffi-dev \
    python-lxml \
\
&& sudo easy_install pip \
&& sudo -H pip install --upgrade ndg-httpsclient #Resolves insecurity warnings in pip
&& sudo -H pip install junos-eznc
```
It's an additional step (using easy_install to install pip) but resolves the issues associated with the distro's archaic version of pip (1.5.4 currently). Using this method, I don't have to worry about explicit version specification of cryptography and it pretty much just works. I don't expect Canonical to change the pip version in trusty unless there's a bug filed, so this could be the best solution. Thoughts?

stacywsmith:
Yes, Nitin and I have been discussing this offline and my current recommendation is to update our Debian/Ubuntu install instructions to have users install pip from source rather than via the Debian package.

I believe the problem is not only that an archaic version of pip is installed, but also that an archaic version of cryptography and setuptools is installed as pre-requisite packages of python-pip.

When PyEZ is installed, even though all of the pip packages correctly specify their dependencies, the setuptools package ends up not getting upgraded and falls back to the archaic version of setuptools installed by the Debian package. This is incompatible with the version of cryptography installed, and produces the error when connecting to a device with PyEZ.

I've tested a slightly different install procedure, but the underlying result is the same because both of our procedures install the pip package for cryptography and setuptools instead of the Debian package.

Here's my procedure:
1) sudo apt-get update
2) sudo apt-get install -y python-dev libxslt1-dev libssl-dev libffi-dev
3) wget https://bootstrap.pypa.io/get-pip.py -O - | sudo python
4a) Install latest released version of PyEZ
      sudo pip install py-junos-eznc
or
4b) Install latest development version of PyEZ
      sudo apt-get install -y git
      sudo pip install git+https://github.com/Juniper/py-junos-eznc.git

ntwrkguru:
I'm OK with that.

PR #537, opened 2016-07-01
Corrected example in documentation

gaddevijay:
Corrected example in documentation


jnpr-community-netdev:

PR #547, opened 2016-07-08
Update rpcmeta.py comments

astiphout:
ansible [documentation](http://junos-ansible-modules.readthedocs.io/en/1.3.1/junos_get_config.html#synopsis) points to jnpr/junos/rpcmeta/get_config; but this did not document how and which options to pass

### DIFF
--- a/lib/jnpr/junos/rpcmeta.py
+++ b/lib/jnpr/junos/rpcmeta.py
@@ -25,9 +25,10 @@ class _RpcMetaExec(object):
         """
         retrieve configuration from the Junos device
 
-        :filter_xml: is options, defines what to retrieve.  if omitted then the entire configuration is returned
+        :filter_xml: XPath expression consisting of XML tag names separated by slashes,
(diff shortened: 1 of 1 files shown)


jnpr-community-netdev:

lib/jnpr/junos/rpcmeta.py
```diff
@@ -25,9 +25,10 @@ def get_config(self, filter_xml=None, options={}):
         """
         retrieve configuration from the Junos device
 
-        :filter_xml: is options, defines what to retrieve.  if omitted then the entire configuration is returned
+        :filter_xml: XPath expression consisting of XML tag names separated by slashes,
+                     defines what to retrieve; when omitted then the entire configuration is returned
 
-        :options: is a dict, creates attributes for the RPC
+        :options: is a dictionary of XML attributes to set within the <get-configuration> RPC.
```
vnitinv: @astiphout Can you please also add example of :options: in the next line for reference. Even better if we can have example for filter_xml

PR #550, opened 2016-07-12
remove select.select on tty to prevent issues with netconf over serial

nembery:
simple fix for netconf over tty with serial connections. Replaces select with plain old read(). Tested with netconf over SSH and serial.


vnitinv:
@nembery HI Nathan, I have tweaked code to use same select.sect mechanism which somehow is faster.
https://github.com/Juniper/py-junos-eznc/pull/551/files#diff-a6148fb119306f8fa78002ea9cb370fbR161

PR #553, opened 2016-07-15
Support to get the  configed ospf interfaces

insqur:
Support to get the  configed ospf interfaces,use to check ospf interfaces configed in OspfInterfaceTable but disappeared in the OspfNeighborTable.


jnpr-community-netdev:

PR #554, opened 2016-07-18
update ospf.yml

insqur:
changing the symbol from the underline to the hyphen.


jnpr-community-netdev:

PR #542, opened 2016-07-04
convert to text for multiple output tags - Juniper/ansible-junos-stdlib#144

stianvi:
##### ISSUE TYPE

 - Bug fix

##### SUMMARY

Fixes Juniper/ansible-junos-stdlib#144
Due to switches returning multiple output-tags when running "show system core-dumps", the xml.etree.elementtree.text won't do.

Tested with both 12.3R9 and 15.1R3 - due to the following rpc-reply containing multiple output tags - we need to use xml.etree.ElementTree.tostring.

Debug output:

```
Jul  4 18:33:13 [NETCONF] - [83745] Incoming: <?xml version="1.0" encoding="UTF-8"?><nc:rpc xmlns:nc="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="urn:uuid:00c66a54-4205-11e6-a4af-0050569f425a"><command>show system core-dumps all-members</command></nc:rpc>]]>]]>
Jul  4 18:33:13 [NETCONF] - [83745] Outgoing: <rpc-reply xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" xmlns:junos="http://xml.juniper.net/junos/15.1R3/junos" xmlns:nc="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="urn:uuid:00c66a54-4205-11e6-a4af-0050569f425a">
Jul  4 18:33:14 [NETCONF] - [83745] Outgoing: <directory-list root-path="/var/crash/*core*" junos:seconds="1467649994" junos:style="verbose">
Jul  4 18:33:14 [NETCONF] - [83745] Outgoing:
Jul  4 18:33:14 [NETCONF] - [83745] Outgoing: <output>
Jul  4 18:33:14 [NETCONF] - [83745] Outgoing: /var/crash/*core*: No such file or directory
Jul  4 18:33:14 [NETCONF] - [83745] Outgoing: </output>
Jul  4 18:33:14 [NETCONF] - [83745] Outgoing: <output>
Jul  4 18:33:14 [NETCONF] - [83745] Outgoing: /var/tmp/*core*: No such file or directory
Jul  4 18:33:14 [NETCONF] - [83745] Outgoing: </output>
Jul  4 18:33:14 [NETCONF] - [83745] Outgoing: <output>
Jul  4 18:33:14 [NETCONF] - [83745] Outgoing: /var/tmp/pics/*core*: No such file or directory
Jul  4 18:33:14 [NETCONF] - [83745] Outgoing: </output>
[...]

PR #564, opened 2016-07-28
dev.cli for multi <output> tags

vnitinv:
resolve Juniper/ansible-junos-stdlib#144 
resolve #542


vnitinv:
```<rpc>
        <get-system-core-dumps format="text">
        </get-system-core-dumps>
    </rpc><rpc-reply xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" xmlns:junos="http://xml.juniper.net/junos/16.1I0/junos">
<output>
fpc0:
--------------------------------------------------------------------------
<output>
/var/crash/*core*: No such file or directory
</output>
-rw-rw----  1 root  field   14172296 Jul 19 09:34 /var/tmp/mgd.core.0.gz
<output>
/var/tmp/pics/*core*: No such file or directory
</output>
<output>
/var/crash/kernel.*: No such file or directory
</output>
<output>
/var/jails/rest-api/tmp/*core*: No such file or directory
</output>
-rw-rw----  1 root  field   14172296 Jul 19 09:34 /tftpboot/corefiles/mgd.core.0.gz
total files: 2
</output>

</rpc-reply>``

PR #569, opened 2016-08-05
structred resource was broken on python 3.x

vnitinv:


vnitinv:
tty_telnet changes inspired from https://github.com/Juniper/py-junos-netconify/pull/48/files

stacywsmith:
@vnitinv The diff in this pull request seems to mix two things. There seems to be the fix for "structred resource was broken on python 3.x" as mentioned in the title, but then there also seems to be a fix for controlling the baud rate when connecting to the console via a Telnet terminal server connection.

I think both changes are OK, but shouldn't be in the same commit.

The changes to write() and rawrite() in lib/jnpr/junos/transport/tty_telnet.py should also be duplicated in the write() and rawrite() methods in lib/jnpr/junos/transport/tty_ssh.py

vnitinv:
@stacy, agree Stacy. That was my mistake. This happens when you use single branch. Next commit becomes part of existing pull request. From next time onward will keep different branch for diff behavior pull requests. I am merging this as you are OK with the changes. Will be careful from now on.

PR #574, opened 2016-08-09
Feature: NSSU and ISSU Upgrade Selection

invaderlucky:
Added an upgrade type functionality that allows the user to run NSSU or ISSU upgrade. The install() method takes an additional parameter called "upType". This parameter can be an S, N, or I, for standard, NSSU, or ISSU. New methods are:

pgkaddNSSU()
      This uses the request_package_nonstop_upgrade.

pgkaddISSU()
      This uses the request_package_in_service_upgrade rpc call.

---
This change is [Reviewable](https://reviewable.io/reviews/juniper/py-junos-eznc/574)

### DIFF
--- a/lib/jnpr/junos/utils/sw.py
+++ b/lib/jnpr/junos/utils/sw.py
@@ -2,7 +2,6 @@
 import hashlib
 import re
 from os import path
-import sys
 
 # 3rd-party modules
 from lxml.builder import E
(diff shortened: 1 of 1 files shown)


jnpr-community-netdev:

vnitinv:
@invaderlucky Looks like the code changes was done on pretty old code and hence the code merge will break other functionalities. 
Also unit-tests need to be added for the new codes. I can help you with this.

vnitinv:
@invaderlucky Dana I see you are calling function pkgupgrade but the given function is not defined anywhere in the class.

vnitinv:
Taken care in #606

PR #578, opened 2016-08-12
replace tounicode with tostring as tounicode is getting depreciated

vnitinv:
help(etree.tounicode)
Help on cython_function_or_method in module lxml.etree:

tounicode(...)
    tounicode(element_or_tree, method="xml", pretty_print=False,
                  with_tail=True, doctype=None)
    
    Serialize an element to the Python unicode representation of its XML
    tree.
    
    :deprecated: use ``tostring(el, encoding='unicode')`` instead.


stacywsmith:
@vnitinv This fix looks good to me and should definitely go into 2.0 to avoid breaking existing code on Python 2.x.

PR #580, opened 2016-08-18
adding timeout to StartShell

vnitinv:


vnitinv:
Example
```python
dev=Device (host=ip, user=username, password=password)
dev.open()
with StartShell(dev, timeout=30) as sh:
    print sh.run("rtsockmon -t")
 
dev.close()
```

vnitinv:
@stacywsmith Can you please review this.

lib/jnpr/junos/utils/start_shell.py
```diff
@@ -112,10 +124,11 @@ def run(self, command, this=_SHELL_PROMPT):
         # run the command and capture the output
         self.send(command)
         got = ''.join(self.wait_for(this))
-
-        # use $? to get the exit code of the command
-        self.send('echo $?')
-        rc = ''.join(self.wait_for(this))
+        rc = ''
+        if re.search(r'(%|>|#)\s?$', got) is not None:
```
stacywsmith: We should only send 'echo $?' if we are at a shell prompt, not a Junos prompt. We shouldn't hardcode the prompt.

Suggest this line should be:
        if re.search(r'{0}\s?$'.format(SHELL_PROMPT), got) is not None:

lib/jnpr/junos/utils/start_shell.py
```diff
@@ -112,10 +124,11 @@ def run(self, command, this=_SHELL_PROMPT):
...
+            # use $? to get the exit code of the command
+            self.send('echo $?')
+            rc = ''.join(self.wait_for(this))
```
stacywsmith: if `this != _SHELL_PROMPT` then this line will end up waiting until the `timeout` value.
Instead, I believe this line should be `rc = ''.join(self.wait_for(_SHELL_PROMPT))`

lib/jnpr/junos/utils/start_shell.py
```diff
@@ -112,10 +124,11 @@ def run(self, command, this=_SHELL_PROMPT):
...
         self.last_ok = True if rc.find('0') > 0 else False
```
stacywsmith: What should the value of `self.last_ok` be when `this != _SHELL_PROMPT` (looking for a non-default prompt)? With the current code, it is set to `False`.

When `this != _SHELL_PROMPT` I believe we should instead set `self.last_ok` to `True` if `this` was found and `False` if `this` was not found.

See the comment above on line 127.

If you agree with my approach, then this line should be part of the conditional block above it.

lib/jnpr/junos/utils/start_shell.py
```diff
@@ -21,28 +22,39 @@ def _ssh_exec(self, command):
...
         :param Device nc: The Device object
+
+        :param int timeout:
+          Timeout value in seconds to wait for expected string/pattern.
         """
         self._nc = nc
+        self.timeout = timeout
 
-    def wait_for(self, this=_SHELL_PROMPT):
+    def wait_for(self, this=_SHELL_PROMPT, timeout=0):
```
stacywsmith: We also need to add a `timeout=0` argument to the `run()` method and pass that `timeout` value through to the `wait_for()` method. Otherwise, you can't specify a unique timeout for a specific command. You just get the timeout specified when the `StartShell` object was created.

lib/jnpr/junos/utils/start_shell.py
```diff
@@ -21,28 +22,39 @@ def _ssh_exec(self, command):
 
     """
 
-    def __init__(self, nc):
+    def __init__(self, nc, timeout=30):
```
stacywsmith: Should `timeout` be a property of the `StartShell` object (as you've done here), or should it come from the `timeout` property of the `Device` object (`nc`), or we could do both by changing this to:
`def __init__(self, nc, timeout=0):`
and my comment on line 35.

lib/jnpr/junos/utils/start_shell.py
```diff
@@ -21,28 +22,39 @@ def _ssh_exec(self, command):
...
         """
         Utility Constructor
 
         :param Device nc: The Device object
+
+        :param int timeout:
+          Timeout value in seconds to wait for expected string/pattern.
         """
         self._nc = nc
+        self.timeout = timeout
```
stacywsmith: Change to:
`self.timeout = timeout if timeout else nc.timeout`

vnitinv:
Thanks @stacywsmith I have started incorporating your feedback.

lib/jnpr/junos/utils/start_shell.py
vnitinv: I disagree with this. nc is just passed to take user credentials. so a user need not to pass nc/dev with a connection opened with device. 
for ex:
```
dev = Device('choc-mx240-k', user='regress', password='MaRtInI', gather_facts=False)
with StartShell(dev) as ss:
    print ss.run('cli show version', '---\(more\s?\d*%?\)---|%')
```
So if you notice dev.open() is not required here.

lib/jnpr/junos/utils/start_shell.py
vnitinv: As mentioned in above case nc.timeout will throw exception

```
  File "xxx", line 17, in <module>
    with StartShell(dev) as ss:
  File "xxx", line 36, in __init__
    self.timeout = timeout or nc.timeout
  File "xxx", line 164, in timeout
    return self._conn.timeout
AttributeError: 'NoneType' object has no attribute 'timeout'
```

device.timeout is only being set when connection is made using dev.open()
Should we change this?

PR #582, opened 2016-08-24
python 2.6 respective changes

vnitinv:
* importlib is requirment for python 2.6
* DeprecationWarning warning for python2.6

### DIFF
--- a/netconify.txt
+++ b/netconify.txt
@@ -1,1 +1,0 @@
-importlib>=1.0.3; python_version == '2.6'
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ addons:
     - libxslt1-dev
 
 install:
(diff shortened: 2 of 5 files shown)


vnitinv:
@stacywsmith Review the changes done in file "lib/jnpr/junos/__init__.py"
inspired from cryptography https://github.com/pyca/cryptography/blob/master/src/cryptography/__init__.py#L21

netconify.txt
```diff
@@ -1 +0,0 @@
-importlib>=1.0.3; python_version == '2.6'
```
stacywsmith: Maybe I'm mis-interpreting the diff output, but it looks like you've removed the only line in the file, but haven't removed the file itself. It doesn't look like the file is needed anymore, so can we remove the file itself?

stacywsmith:
@vnitinv other than my comment about removing the netconify.txt file completely, this pull request looks good to me.

PR #585, opened 2016-09-01
Added Dockerfile to enable automated image builds on project commits

ntwrkguru:


vnitinv:
@dgarros Can you please review this Pull request.

dgarros:
Looks great !! I'm very happy to see us moving in this direction, thanks @ntwrkguru 
Once it's merged, I'll update docker hub to build it automatically

dgarros:
Hi @ntwrkguru, @vnitinv 

I changed the source repository on Dockerhub 
For each commit on this repo, a new container will be automatically created
And a new docker tag will be created for each tag on github

I also created and uploaded docker images for versions `1.3.1` and `2.0.0`
https://hub.docker.com/r/juniper/pyez/tags/
I would recommend most people to use these instead of `latest`

In the process I found some missing part in the Dockerfile, so I fixed them.

Issue #587, opened 2016-09-02
Config.lock() return exception when normalize is on


jnpr-raylam:
>>> dev=Device('172.27.62.164',user='root',passwd='lab123',normalize=True)
>>> dev.open()
Device(172.27.62.164)
>>> cu=Config(dev)
>>> cu.lock()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/root/pyez2.0/lib/python2.7/site-packages/jnpr/junos/utils/config.py", line 467, in lock
    raise LockError(rsp=JXML.remove_namespaces(err.xml))
AttributeError: 'exceptions.AttributeError' object has no attribute 'xml'

If normalize is not enabled, no problem
This problem is introduced in pyez 2.0

PR #591, opened 2016-09-07
print dev for Console conn was printing object not Device(....)

vnitinv:
```python
>>> from jnpr.junos import Device
>>> dev = Device('xxx', user='xxx', password='xxx', mode='telnet', port='23', gather_facts=True)
>>> dev.open()
<jnpr.junos.console.Console object at 0x7f9bd42cfa90>
```
After Fix
```python
>>> from jnpr.junos import Device
>>> dev = Device('xxx', user='xxx', password='xxx', mode='telnet', port='23', gather_facts=True)
>>> dev.open()
Device(xxx)
```

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -480,6 +480,13 @@ class _Connection(object):
                               RuntimeWarning)
                 return
 
+    # -----------------------------------------------------------------------
+    # OVERLOADS
+    # -----------------------------------------------------------------------
+
+    def __repr__(self):
+        return "Device(%s)" % self.hostname
+
 
 class Device(_Connection):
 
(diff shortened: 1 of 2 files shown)


lib/jnpr/junos/device.py
```diff
@@ -919,7 +919,7 @@ def execute(self, rpc_cmd, **kvargs):
             #    protocol: operation-failed
             #    error: device asdf not found
             # </rpc-reply>
-            if rpc_rsp_e.text is not None and rpc_rsp_e.text.strip() is not '':
+            if rpc_rsp_e.text.strip() is not '':
```
stacywsmith: This line looks like it accidentally rolled back the commit for issue #587 (fixed in pull request #590).

vnitinv: @stacywsmith Yes, Will fix this. Thanks a lot Stacy

PR #603, opened 2016-09-23
Enhance warning message from the `cli()` method.

stacywsmith:
When the `cli()` method was called without `warning=False`, the following warning was displayed:
```
CLI command is for debug use only!
```

This warning has now been enhanced to inform users of the equivalent RPC method. The new warning is:
```
CLI command is for debug use only!
Instead of:
cli('show interfaces terse lo0.0')
Use:
rpc.get_interface_information(terse=True, interface_name='lo0.0')
```

In addition, the warning will be suppressed if there is no RPC equivalent.


vnitinv:
@stacywsmith You will have to add the corresponding UT too. As you can see code coverage went down by almost 1%

stacywsmith:
@vnitinv I now think this is ready. Can you please review?

vnitinv:
functionality wise looks good to me @stacywsmith But I can see few pep8 violation. 
Can you check with below command
```
pep8 device.py
```
To automatically resolve these issue we can use autopep8 tool.
```
autopep8 --in-place --aggressive --recursive device.py
```

PR #606, opened 2016-09-27
Issu sw install

vnitinv:


lib/jnpr/junos/utils/sw.py
```diff
@@ -170,30 +170,71 @@ def pkgadd(self, remote_package, **kvargs):
...
+        return self._parse_pkgadd_response(rsp)
+
+    def _parse_pkgadd_response(self, rsp):
         got = rsp.getparent()
         rc = int(got.findtext('package-result').strip())
-
-        # return True if rc == 0 else got.findtext('output').strip()
-        return True if rc == 0 else False
+        if rc != 0:
+            self.log("software pkgadd message: %s" % got.findtext('output'))
```
stacywsmith: Suggest unconditionally logging this message, instead of just logging if there's a failure.

lib/jnpr/junos/utils/sw.py
```diff
@@ -430,11 +494,20 @@ def _progress(report):
...
-                v_ok = self.validate(remote_package, dev_timeout=timeout)
-                if v_ok is not True:
-                    return v_ok  # will be the string of output
+                v_ok = self.validate(remote_package, issu, dev_timeout=timeout)
 
-            if self._multi_RE is False:
+                if v_ok is not True:
+                    return v_ok
+
+            if issu is True:
```
stacywsmith: ISSU and NSSU only make sense when `self._multi_RE is True`. Currently the code tries them even if `self._multi_RE is False`. Maybe this is OK because the RPCs themselves will likely fail in this case. However, I'm wondering if it makes more sense for us to only attempt ISSU/NSSU when `self._multi_RE is True`.

lib/jnpr/junos/utils/sw.py
vnitinv: Taken care. Thanks

lib/jnpr/junos/utils/sw.py
```diff
@@ -430,11 +494,20 @@ def _progress(report):
                 _progress(
                     "validating software against current config,"
                     " please be patient ...")
-                v_ok = self.validate(remote_package, dev_timeout=timeout)
-                if v_ok is not True:
-                    return v_ok  # will be the string of output
+                v_ok = self.validate(remote_package, issu, dev_timeout=timeout)
```
vnitinv: I think we should not change this. Timeout in general meant for per rpc. So it should be individual for each function call internally too, which is happening as of today.

lib/jnpr/junos/utils/sw.py
vnitinv: I am adding this. The plus point I saw by not putting that condition was even when the device was connected using gather_facts as False or facts gathering failed, this would work perfectly fine. I have added the check for now.

vnitinv:
sample example:
```python
import time
from jnpr.junos import Device
from pprint import pprint
from jnpr.junos.utils.sw import SW
from jnpr.junos.utils.config import Config
from lxml import etree

with Device(host='xxxx’, user=’xxxx’, password=’xxx’) as dev:
    pprint(dev.facts)
    facts = dev.facts
    # dev.facts_refresh(exception_on_failure=True)
    if not facts['version_RE0'] == facts['version_RE1']:
        raise Exception('Not compatible for ISSU upgrade')

    # Inspired from
    # http://www.juniper.net/documentation/en_US/junos16.1/topics/task/installation/issu-upgrading.html#test1
    set_commands = """
    set chassis redundancy graceful-switchover
    set routing-options nonstop-routing
    set system commit synchronize"""
    # this would take care of lock and unlock
    with Config(dev, mode='exclusive') as cu:
        cu.load(set_commands)
        cu.commit()

    # Verify that NSR is configured on the master Routing Engine (re0) by using the show task replication command.
    op = dev.rpc.get_routing_task_replication_state()
    if not op.findtext('task-gres-state') == 'Enabled' and not op.findtext('') == 'Master':
        raise Exception('"show task replication" is not as per ISSU requirement')

    # Verify that GRES is enabled on the backup Routing Engine (re1) by using the show system switchover command.
    # As I know backup name, below code will work, otherwise I can use startShell to go from master to backup
    with Device(host='xxxx1’, user=’xxxx’, password=’xxx’, gather_facts=False) as dev1:
        op = dev1.rpc.get_switchover_information()
        if op.findtext('switchover-state').strip().lower() != 'on':
            raise Exception('"switchover-state is not On')

    # Adjusting Timers and Changing Feature-Specific Configuration
    # I am not handling this. Do we need to check this too?

    sw = SW(dev)

    try:
        ok = sw.install(package='junos-install-mx-x86-64-16.1-20160925.0.tgz', issu=True,
                  no_copy=True, validate=True, progress=True)
        if ok is True:
            # below steps is only required you we want to reboot the old master re only
            sw.reboot(all_re=False)
            # device might be pingable for few seconds before it goes for reboot.
            time.sleep(30)
            if dev.probe(1200, 5):
                print "old master is rebooted and is now up"
                dev.open(gather_facts=False)
                #time.sleep(300) # without sleep we get message "Not ready for mastership switch, try after 239 secs."
                print dev.cli("request chassis routing-engine master acquire", warning=False)
                print "Device is UP & ISSU upgrade Done"
            else:
                print "old master is not pingable even after 1200 sec after reboot"
        else:
            "sw.install returned False"
    except Exception as ex:
        print "ERROR: %s" % ex.message
```

PR #614, opened 2016-10-19
removing the hard check of serial number if present then only assigned

Jagrut:
`Device(host=ip, user='user', password="xxxxxx", os="JUNOS", connect_mode="ssh")`

In above Simple ssh is failing because device doesn't have a serial number. In the juniper too many device doesn't have a serial number. In this pull request hard checking of serial number is removed. assign only if it presents.

### DIFF
--- a/lib/jnpr/junos/facts/chassis.py
+++ b/lib/jnpr/junos/facts/chassis.py
@@ -43,10 +43,11 @@ def facts_chassis(junos, facts):
 
     facts['model'] = x_ch.findtext('description')
 
-    try:
+    #try:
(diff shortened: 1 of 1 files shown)


jnpr-community-netdev:

stacywsmith:
@Jagrut Closing this pull request. I've got #615 to address this in a more generic and cleaner way. Will have @vnitinv review and commit #615.

Jagrut:
@stacywsmith and @vnitinv  . I also want to contribute. I want to fix this issue.  @vnitinv  can review my changes if you ppl are agree.

PR #615, opened 2016-10-20
Address issue with fact gathering stopping when it encounters a problem.

stacywsmith:
Cleaned up the chassis related fact gathering to
report "UNKNOWN" for model and serial number if it has a problem.
Addresses #614 in a cleaner way.


vnitinv:
Jenkins: ok to test

Issue #616, opened 2016-10-21
Pipe raw CLI output for commands without RPC equiv


mirceaulinic:
As per our discussion in the forums, I am going to add the facility to pipe the output.

For some obscure reasons, although the device returns all these details (or at lease this is what logs show), ncclient does not receive that.

Logs from device:

```
Oct 21 00:23:29 [1709] Incoming: <?xml version="1.0" encoding="UTF-8"?><nc:rpc xmlns:nc="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="urn:uuid:4fbc76f2-2e45-40cd-9010-2845524715aa"><command format="text">show configuration | display set | match napalm</command></nc:rpc>]]>]]>
Oct 21 00:23:29 [1709] Outgoing: <rpc-reply xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" xmlns:junos="http://xml.juniper.net/junos/12.1X47/junos" xmlns:nc="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="urn:uuid:4fbc76f2-2e45-40cd-9010-2845524715aa">
Oct 21 00:23:29 [1709] Outgoing: <pipe>
Oct 21 00:23:29 [1709] Outgoing: <more-match-string>
Oct 21 00:23:29 [1709] Outgoing: napalm
Oct 21 00:23:29 [1709] Outgoing: </more-match-string>
Oct 21 00:23:29 [1709] Outgoing: </pipe>
Oct 21 00:23:29 [1709] Outgoing: <configuration-information>
Oct 21 00:23:29 [1709] Outgoing: <configuration-output>
Oct 21 00:23:29 [1709] Outgoing: set version 12.1X47-D20.7
Oct 21 00:23:29 [1709] Outgoing: set system host-name vsrx
Oct 21 00:23:29 [1709] Outgoing: set system root-authentication encrypted-password "$1$5MhDFyrI$NBBMndW1POqbN.0QEA4z0."
[...]

PR #617, opened 2016-10-21
PoC for piping the raw CLI output

mirceaulinic:
Opening this PR for discussion around piping raw CLI outputs #616 

Implemented only match and count so far. In case the model proposed looks good, we can move on and implement the others:

```
Possible completions:
  count                Count occurrences
  display              Show additional kinds of information
  except               Show only text that does not match a pattern
  find                 Search for first occurrence of pattern
  hold                 Hold text without exiting the --More-- prompt
  last                 Display end of output only
  match                Show only text that matches a pattern
  no-more              Don't paginate output
  refresh              Refresh a continuous display of the command
  request              Make system-level requests
  resolve              Resolve IP addresses
  save                 Save output text to file
  trim                 Trim specified number of columns from start of line
```

Of course, except things such as no-more, as anyway it is provided the complete output, or refresh.

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -4,6 +4,7 @@ import types
 import platform
 import warnings
 import traceback
+from collections import OrderedDict
 
 # stdlib, in support of the the 'probe' method
 import socket
@@ -469,6 +470,49 @@ class _Connection(object):
             or ``| count``, etc.  The only value use of the "pipe" is for the
             ``| display xml rpc`` as noted above.
         """
+
+        def _count(txt, none):
+            count = len(txt.splitlines())
+            return 'Count: {count} lines'.format(count=count)
+            # will return the exact output, as Junos displays
+            # e.g.:
+            # > show system processes extensive | match root | count
+            # Count: 113 lines
+
+        def _match(txt, pattern):
+            rgx = '^.*({pattern}).*$'.format(pattern=pattern)
+            matched = [
+                line for line in txt.splitlines()
+                if re.search(rgx, line)
+            ]
+            return '\n'.join(matched)
+
+        def _filter(txt, command):
+            if not txt:
+                return txt
+
+            _OF_MAP = OrderedDict()
+            _OF_MAP['match'] = _match
+            _OF_MAP['count'] = _count
+            # the operations order matter in this case!
+
+            exploded_command = command.split('|')
+            pipe_oper_args = {}
+            for pipe in exploded_command[1:]:
+                exploded_pipe = pipe.split()
+                pipe_oper = exploded_pipe[0]  # always there
+                pipe_args = ''.join(exploded_pipe[1:2])
+                # will not throw error when there's no arg
+                pipe_oper_args[pipe_oper] = pipe_args
+
+            for oper in _OF_MAP.keys():
+                # to make sure the operation sequence is correct
+                if oper not in pipe_oper_args.keys():
+                    continue
+                txt = _OF_MAP[oper](txt, pipe_oper_args[oper])
+
+            return txt
+
         if 'display xml rpc' not in command and warning is True:
             # Get the equivalent rpc metamethod
             rpc_string = self.cli_to_rpc_string(command)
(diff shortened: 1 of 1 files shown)


jnpr-community-netdev:

mirceaulinic:
@jnpr-community-netdev are you sure this library is still maintained? :P

Seriously now: @vnitinv @stacywsmith any thoughts on this one-month-old PR?

vnitinv:
@mirceaulinic Sorry for late response. Was on PTO for whole Nov. We are evaluating option to have support for pipe '|' over netconf from backend itself. Supporting cli kind of output by PyEZ will be a big burden, as with different device types output keeps changing. I hope you do understand.

mirceaulinic:
Hi @vnitinv - thanks for looking into this!

>  I hope you do understand.

Sure, I understand, no problem; happy to close this PR if anything better in place.

> We are evaluating option to have support for pipe '|' over netconf from backend itself.

That's very good news! It makes much more sense to be like that, however I have two questions:

1. Is this going to be supported on older versions (I would assume on Junos 12 or 13 we can't expect any)?
2. How about the other options, e.g.: ```count```, ```match```, ```except``` etc.? Will they be supported as well?

Cheers!

mirceaulinic:
@vnitinv is this still on your radar? My hunch is that the solution you have exposed will never be available in older releases, most probably it will require Junos >= 16 or even >=17, which is not a comfortable constraint given that we are able to provide a workaround, without breaking the existing behaviour.

This feature is based on the feedback of many engineers working in production environments. I know the theory, but, as always, in the real-world there are more specific needs. And filtering the output is one of them.

@stacywsmith may I have also your input on this?

Thanks,
Mircea

vnitinv:

attila123:
@mirceaulinic this was more than a year ago. Are you still willing to finish this patch?
Seems like Travis reported some error.

mirceaulinic:
Hi @attila123 - I think I'm waiting here for feedback from the maintainers... is Juniper still against doing this thing?

lstoor:
Hi,
Is this thread still active? I have also run into "|" related problems.
Br,
Lakhavir

attila123:
Looks like only "| display xml rpc" piping is supported, see method documentation from https://github.com/Juniper/py-junos-eznc/blob/5b6c3a1f2c02e11afd7e056eaeb99ef50759a3b7/lib/jnpr/junos/device.py#L661 onwards.

ntwrkguru:
> Hi @attila123 - I think I'm waiting here for feedback from the maintainers... is Juniper still against doing this thing?

A year late, but I am personally opposed to this as programmatically invoking CLI is anti-pattern when there are other mechanisms to retrieve and parse the data within the NETCONF system itself. What I mean is that the CLI simply passes a NETCONF command, and any piping is doing some processing prior to returning the data to the user. `display xml rpc` is the proper exception, IMO, as it allows for retrieving the RPC equivalent of a given CLI command. Several of the arguments to the pipe don't make sense for programmatic access; e.g. refresh, resolve, hold, etc.

All in all, passing CLI over NETCONF is a bit of a hack, even if it is used a lot. However, this is just my 2c and I don't represent all of Juniper or even this project. :-)

attila123:
@ntwrkguru In our use case this cli can be executed remotely by operations people in troubleshooting situations. This is exactly what the cli is meant to be used for: debugging, troubleshooting. But currently cli is there, but in a crippled version, which limits the possibilities by not allowing most piping options. Looks like now I need to implement my own cli client with direct ssh (e.g. with paramiko library).

ntwrkguru:
> @ntwrkguru In our use case this cli can be executed remotely by operations people in troubleshooting situations. This is exactly what the cli is meant to be used for: debugging, troubleshooting. But currently cli is there, but in a crippled version, which limits the possibilities by not allowing most piping options. Looks like now I need to implement my own cli client with direct ssh (e.g. with paramiko library).
> Update: I see now in the code that it actually does not execute the cli command as cli command, but first coverts it to rpc, which can explain why devs do not want to implement all possible pipe operations.

And one layer down, mgd actually converts CLI to RPC on the box as well. :-) To your point, this is the pain of cultural transformation from a box-driven paradigm to an automated one.

rahkumar651991:
Based on the above discussions, the rpc based approach doesn't sync well with the cli pipe based implementation. 
Closing this pull request.

PR #621, opened 2016-11-05
Minor typos fixed in RuntimeError exception message and in comments

pklimai:
I tried to make PyEZ work with my devices via console (terminal server) and had some problems. I haven't been able to localize the problem yet, but while browsing the sources found some typos and fixed them in this commit. Hopefully this improves overall code readability.

I wonder if you are interested in such type of contributions.


jnpr-community-netdev:

vnitinv:
Thanks @pklimai

PR #622, opened 2016-11-10
Added console_has_banner parameter

pklimai:
Hi

Basically, problem that I am solving with this proposed modification is described [here](http://stackoverflow.com/questions/40493376/how-to-make-junos-pyez-work-via-console-port)

In the existing PyEZ version, in comments, it is suggested to uncomment couple of lines in _ev_tty_nologin() function in case "we're in a hung state" and "console connection has a banner or warning". Really, uncommenting the lines solved my problem but this is not what users are supposed to do (modify library code).

So I propose to add the parameter console_has_banner to the Console (also Device) class constructor, allowing to make this switch more easily for users.

With this parameter enabled, this script now works much better:

```
with Device(host='X.X.X.X', user='XXX', password='XXXXXX', mode='telnet', 
        port='2014', gather_facts=True, console_has_banner=True) as dev:
    print(dev.facts)
    print(dev.cli("show version", format='text', warning=False))
```

### DIFF
--- a/lib/jnpr/junos/console.py
+++ b/lib/jnpr/junos/console.py
@@ -75,6 +75,11 @@ class Console(_Connection):
             *OPTIONAL* default is ``False``.  If ``False`` then the
             facts are not gathered on call to :meth:`open`
 
+        :param bool console_has_banner:
+            *OPTIONAL* default is ``False``.  If ``False`` then in case of a
+            hung state, <close-session/> rpc is sent to the console.
+            If ``True``, after sleep(5), a new-line is sent
+
         """
 
         # ----------------------------------------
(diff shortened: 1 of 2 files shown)


jnpr-community-netdev:

stacywsmith:
@pklimai In the long run, I think we need a better state machine to handle the console login/logout properly even when banner's exist. For now, I'll merge this request as it seems like it addresses the immediate issue and shouldn't cause any regressions since you added it as a non-default parameter. Thanks for submitting the pull request!

pklimai:
Thanks Stacy, very much appreciated. 
I will also update the question on StackOverflow accordingly.

PR #623, opened 2016-11-15
Add CentOS Support to install instructions

kkirsche:
Add `openssl-devel` dependency for CentOS 7 and update that this works with CentOS via filename and listing


stacywsmith:
Thanks @kkirsche

kkirsche:
Of course. Happy to try to help others out with this kind of thing

PR #628, opened 2016-11-21
Lookup user from ~/.ssh/config

vincentbernat:
Due to a change in commit b03e611ef4eb, the user wasn't extracted from
user (or specified) SSH configuration. Add back the call to parse SSH
configuration.

Also, don't use the parsed private keyfile as it will be done
automatically by the underlying Paramiko. Using the parsed key doesn't
work, either because this is a list and not a string or because it is
encrypted and I am using an agent.

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -833,6 +833,7 @@ class Device(_Connection):
             self._conf_ssh_private_key_file = None
             # user can get updated by ssh_config
             self._ssh_config = kvargs.get('ssh_config')
+            self._sshconf_lkup()
             # but if user or private key is explicit from call, then use it.
             self._auth_user = kvargs.get('user') or self._conf_auth_user or \
                 self._auth_user


ydnath:
@stacywsmith and @vnitinv to review

lib/jnpr/junos/device.py
```diff
@@ -833,11 +833,11 @@ def __init__(self, *vargs, **kvargs):
             self._conf_ssh_private_key_file = None
             # user can get updated by ssh_config
             self._ssh_config = kvargs.get('ssh_config')
+            self._sshconf_lkup()
```
vnitinv: I agree adding this line which will override self._conf_ssh_private_key_file None value set in line 833 if that is configured in ssh config file.

lib/jnpr/junos/device.py
```diff
@@ -833,11 +833,11 @@ def __init__(self, *vargs, **kvargs):
...
             # but if user or private key is explicit from call, then use it.
             self._auth_user = kvargs.get('user') or self._conf_auth_user or \
                 self._auth_user
-            self._ssh_private_key_file = kvargs.get('ssh_private_key_file') \
-                or self._conf_ssh_private_key_file
+            self._ssh_private_key_file = kvargs.get('ssh_private_key_file')
```
vnitinv: why "or self._conf_ssh_private_key_file" was removed. If user has configured private key file in ssh config file, by removing the code the value will be set to None only.

vincentbernat: Yes, you are right. I don't know why I removed it. I have amended the commit.

PR #629, opened 2016-11-28
Added directory_usage to utils

pklimai:
In PyEZ filesystem utils we were missing a method to see the size of the particular directory (together with all subdirectories). I added directory_usage util to do that.

Unit test for this util added as well.

### DIFF
--- a/lib/jnpr/junos/utils/fs.py
+++ b/lib/jnpr/junos/utils/fs.py
@@ -20,6 +20,7 @@ class FS(Util):
     * :meth:`rmdir`: remove a directory
     * :meth:`stat`: return file/dir information
     * :meth:`storage_usage`: return storage usage
+    * :meth:`directory_usage`: return directory usage in bytes
     * :meth:`storage_cleanup`: perform storage storage_cleanup
     * :meth:`storage_cleanup_check`: returns a list of files to remove at cleanup
     * :meth:`symlink`: create a symlink
@@ -264,6 +265,35 @@ class FS(Util):
         return dict((_name(fs), _decode(fs)) for fs in rsp.xpath('filesystem'))
 
     # -------------------------------------------------------------------------
+    # directory_usage - filesystem directory usage
+    # -------------------------------------------------------------------------
+
+    def directory_usage(self, path="."):
+        """
+        Returns the directory usage, similar to the unix "du" command.
+
+        :returns: approximate directory usage, including subdirectories, in bytes
+        """
+        rsp = self._dev.rpc.get_directory_usage_information(path=path, depth="0")
+
+        used_space = rsp.findtext("directory/used-space")
+
+        if used_space is None:
+            return None
+
+        used_space = used_space.strip()
+
+        multiplier = {
+            'B': 1,
+            'K': 1024,
+            'M': 1024**2,
+            'G': 1024**3,
+            'T': 1024**4,
+        }
+
+        return int(float(used_space[:-1]) * multiplier[used_space[-1]])
+
+    # -------------------------------------------------------------------------
     ### storage_cleanup_check, storage_cleanip
     # -------------------------------------------------------------------------
 
@@ -388,8 +418,8 @@ class FS(Util):
         return rsp.text
 
     # -------------------------------------------------------------------------
-    # !!!!! methods that use SSH shell commands, requires that the user
-    # !!!!! has 'start shell' priveldges
+    # !!!!! methods that use SSH shell commands, require that the user
+    # !!!!! has 'start shell' privileges
     # -------------------------------------------------------------------------
 
     def _ssh_exec(self, command):
--- a/lib/jnpr/junos/utils/fs.py
+++ b/lib/jnpr/junos/utils/fs.py
@@ -280,22 +280,25 @@ class FS(Util):
 
         rsp = self._dev.rpc.get_directory_usage_information(path=path, depth=str(depth))
 
-        if rsp.findtext("directory/used-space") is None:
-            # Likely, no such directory
-            raise RpcError(rsp=rsp)
-
-        dirs = rsp.xpath("//directory")
         result = {}
 
-        for directory in dirs:
+        for directory in rsp.findall(".//directory"):
             dir_name = directory.findtext("directory-name").strip()
-            dir_size = directory.findtext("used-space").strip()
-            dir_blocks = int(directory.find("used-space").get("used-blocks").strip())
-            result[dir_name] = {
-                "size": dir_size,
-                "blocks": dir_blocks,
-                "bytes": dir_blocks * BLOCK_SIZE,
-            }
+            if dir_name is None:
+                raise RpcError(rsp=rsp)
+
+            used_space = directory.find('used-space')
+            if used_space is not None:
+                dir_size = used_space.text.strip()
+                dir_blocks = used_space.get('used-blocks')
+                if dir_blocks is not None:
+                    dir_blocks = int(dir_blocks)
+                    dir_bytes = dir_blocks * BLOCK_SIZE
+                    result[dir_name] = {
+                        "size": dir_size,
+                        "blocks": dir_blocks,
+                        "bytes": dir_bytes,
+                    }
 
         return result
 
--- a/lib/jnpr/junos/utils/fs.py
+++ b/lib/jnpr/junos/utils/fs.py
@@ -3,6 +3,8 @@ from lxml.builder import E
 from jnpr.junos.utils.util import Util
 from jnpr.junos.utils.start_shell import StartShell
 
+from jnpr.junos.exception import RpcError
+
 
 class FS(Util):
     """
@@ -20,7 +22,7 @@ class FS(Util):
     * :meth:`rmdir`: remove a directory
     * :meth:`stat`: return file/dir information
     * :meth:`storage_usage`: return storage usage
-    * :meth:`directory_usage`: return directory usage in bytes
+    * :meth:`directory_usage`: return directory usage
     * :meth:`storage_cleanup`: perform storage storage_cleanup
     * :meth:`storage_cleanup_check`: returns a list of files to remove at cleanup
     * :meth:`symlink`: create a symlink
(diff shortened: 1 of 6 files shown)


lib/jnpr/junos/utils/fs.py
```diff
@@ -264,6 +265,35 @@ def _decode(fs):
         return dict((_name(fs), _decode(fs)) for fs in rsp.xpath('filesystem'))
 
     # -------------------------------------------------------------------------
+    # directory_usage - filesystem directory usage
+    # -------------------------------------------------------------------------
+
+    def directory_usage(self, path="."):
```
stacywsmith: Suggest adding a `depth=0` argument to the method.

lib/jnpr/junos/utils/fs.py
```diff
@@ -264,6 +265,35 @@ def _decode(fs):
...
+        """
+        Returns the directory usage, similar to the unix "du" command.
+
+        :returns: approximate directory usage, including subdirectories, in bytes
```
stacywsmith: Rather than returning a single integer, I suggest returning a dictionary similar to the return value of `storage_usage()`. The first level key would be directory name. The second level keys could be `blocks`, `bytes`, and `size` (where `size` would be the 'human-readable' value with units).

Returning this type of data structure allows the support of the `depth > 0` situation and allows returning the values in multiple formats so the user can easily access their preferred format without having to perform conversion.

lib/jnpr/junos/utils/fs.py
```diff
@@ -264,6 +265,35 @@ def _decode(fs):
...
+        """
+        rsp = self._dev.rpc.get_directory_usage_information(path=path, depth="0")
```
stacywsmith: Assuming `depth` is now an argument method per comment on line 271, change `depth="0"` to `depth=str(depth)`.

lib/jnpr/junos/utils/fs.py
```diff
@@ -264,6 +265,35 @@ def _decode(fs):
...
+
+        used_space = rsp.findtext("directory/used-space")
+
+        if used_space is None:
+            return None
+
+        used_space = used_space.strip()
+
+        multiplier = {
```
stacywsmith: Calculate the `bytes` value using the number of blocks (the `used-blocks` attribute of the `<used-space>` element. Then it's simply: `bytes = blocks * BLOCK_SIZE`. You can define `BLOCK_SIZE = 512` at the beginning of the method.

lib/jnpr/junos/utils/fs.py
stacywsmith: If you implement my comments, then "approximate" can be quantified as 512-byte precision.

lib/jnpr/junos/utils/fs.py
stacywsmith: There's a possibility this line will raise any of the exceptions raised when invoking an RPC. That's fine, but we might want to document it.

lib/jnpr/junos/utils/fs.py
```diff
@@ -264,6 +265,35 @@ def _decode(fs):
...
+        Returns the directory usage, similar to the unix "du" command.
+
+        :returns: approximate directory usage, including subdirectories, in bytes
+        """
+        rsp = self._dev.rpc.get_directory_usage_information(path=path, depth="0")
+
+        used_space = rsp.findtext("directory/used-space")
+
+        if used_space is None:
+            return None
```
stacywsmith: I'm not sure returning `None` makes sense for a missing `<used-space>` element. Especially if you implement my recommendation on changing the return value to a `dict`. Instead, I suggest raising a `RpcError` exception.

pklimai:
Hi Stacy,

Thanks for your excellent ideas, I will try to implement it. 

The only thing I'm not sure is why you want to make "depth" the method parameter. I just want directory_usage method to return usage for directory provided in the "path".

So for example with depth=0 the return XML can be

`lab@R1> show system directory-usage depth 0 | display xml   
<rpc-reply xmlns:junos="http://xml.juniper.net/junos/12.1X47/junos">
    <directory-usage-information>
        <directory>
            <directory-name>/cf/var/home/lab</directory-name>
            <used-space used-blocks="4356">
                2.1M
            </used-space>
        </directory>
    </directory-usage-information>
    <cli>
        <banner></banner>
    </cli>
</rpc-reply>`

while the same for depth=1

`lab@R1> show system directory-usage depth 1 | display xml    
<rpc-reply xmlns:junos="http://xml.juniper.net/junos/12.1X47/junos">
    <directory-usage-information>
        <directory>
            <directory-name>/cf/var/home/lab</directory-name>
            <directory>
                <directory-name>/cf/var/home/lab/ajer</directory-name>
                <used-space used-blocks="16">
                    8.0K
                </used-space>
            </directory>
            <directory>
                <directory-name>/cf/var/home/lab/.ssh</directory-name>
                <used-space used-blocks="8">
                    4.0K
                </used-space>
            </directory>
            <directory>
                <directory-name>/cf/var/home/lab/test</directory-name>
                <used-space used-blocks="3984">
                    1.9M
                </used-space>
            </directory>
            <directory>
                <directory-name>/cf/var/home/lab/ajer-trunks</directory-name>
                <used-space used-blocks="168">
                    84K
                </used-space>
            </directory>
            <used-space used-blocks="4356">
                2.1M
            </used-space>
        </directory>
    </directory-usage-information>
    <cli>
        <banner></banner>
    </cli>
</rpc-reply>`

so there is a lot more information here, but all that is not needed - we still only need "directory/used-space" which is 2.1M anyway. So calling with other depths just consumes extra resources. Please tell me what you think.

- Peter

stacywsmith:
I understand your use case for the directory method, but I can also envision other use cases which can be handled by adding a `depth` argument. Really it's the same reason the Unix `du` command takes a depth argument.

Here's a quick example with `du`.

Sometimes I want the space consumed by all the home directories on the system:
```
% du -h -d 0 /var/home
 20K	/var/home
```

Other times I'd like a single call to tell me the space consumed by each user's home directory:
```
% du -h -d 1 /var/home
4.0K	/var/home/regress
4.0K	/var/home/fregress
4.0K	/var/home/remote
4.0K	/var/home/user
 20K	/var/home
```

Yes, you could invoke `directory_usage()` once for each subdirectory, but it's more efficient and easier to only have to invoke it once.

pklimai:
Thanks Stacy, I think you are right and it is better to have a more general method.

Just pushed the changes, please review.

lib/jnpr/junos/utils/fs.py
```diff
@@ -268,30 +270,34 @@ def _decode(fs):
...
+        result = {}
 
-        multiplier = {
-            'B': 1,
-            'K': 1024,
-            'M': 1024**2,
-            'G': 1024**3,
-            'T': 1024**4,
-        }
+        for directory in dirs:
```
stacywsmith: Change to `for directory in rsp.findall('//directory'):`

lib/jnpr/junos/utils/fs.py
```diff
@@ -268,30 +270,34 @@ def _decode(fs):
...
+        rsp = self._dev.rpc.get_directory_usage_information(path=path, depth=str(depth))
 
-        if used_space is None:
-            return None
+        if rsp.findtext("directory/used-space") is None:
+            # Likely, no such directory
+            raise RpcError(rsp=rsp)
 
-        used_space = used_space.strip()
+        dirs = rsp.xpath("//directory")
```
stacywsmith: Delete this line. Use `findall()` on line 290 instead.

lib/jnpr/junos/utils/fs.py
```diff
@@ -268,30 +270,34 @@ def _decode(fs):
...
         Returns the directory usage, similar to the unix "du" command.
 
-        :returns: approximate directory usage, including subdirectories, in bytes
+        :returns: dict of directory usage, including subdirectories if depth > 0
         """
-        rsp = self._dev.rpc.get_directory_usage_information(path=path, depth="0")
+        BLOCK_SIZE = 512
 
-        used_space = rsp.findtext("directory/used-space")
+        rsp = self._dev.rpc.get_directory_usage_information(path=path, depth=str(depth))
```
stacywsmith: Add `normalize=True` argument. So, line becomes:
`rsp = self._dev.rpc.get_directory_usage_information(path=path, depth=str(depth), normalize=True)`
This avoids having to call `strip()` on lines 291-293.

lib/jnpr/junos/utils/fs.py
```diff
@@ -268,30 +270,34 @@ def _decode(fs):
...
 
-        if used_space is None:
-            return None
+        if rsp.findtext("directory/used-space") is None:
+            # Likely, no such directory
+            raise RpcError(rsp=rsp)
```
stacywsmith: I'm sorry. I know I suggested this change, but now I'm rethinking it.
This would only raise an `RpcError` if there was no `<directory>` element in the response that contained a `<used-space>` element. If there were multiple `<directory>` elements in the response, but only some of them were missing the `<used-space>` element, it wouldn't address that situation.

Now that you've modified to return a dictionary, we probably need to handle the exceptional conditions of missing elements within the loop over each `<directory>` element. I suggest removing lines 283-285.

lib/jnpr/junos/utils/fs.py
```diff
@@ -268,30 +270,34 @@ def _decode(fs):
...
 
-        multiplier = {
-            'B': 1,
-            'K': 1024,
-            'M': 1024**2,
-            'G': 1024**3,
-            'T': 1024**4,
-        }
+        for directory in dirs:
+            dir_name = directory.findtext("directory-name").strip()
```
stacywsmith: Remove `.strip()` from the end. If `findtext()` doesn't find a matching element, it returns `None`. Calling `strip()` on a `None` value would cause an exception to be raised. We can avoid this situation by implementing the comment on line 281.

stacywsmith: Need to check that the `<directory-name>` element was actually found. Otherwise, we will end up trying to add a key with a value of `None` to the dict. I suggest adding the following after line 291:
```
if dir_name is None:
    raise RpcError(rsp=rsp)
```

lib/jnpr/junos/utils/fs.py
```diff
@@ -268,30 +270,34 @@ def _decode(fs):
...
+            dir_size = directory.findtext("used-space").strip()
+            dir_blocks = int(directory.find("used-space").get("used-blocks").strip())
+            result[dir_name] = {
+                "size": dir_size,
+                "blocks": dir_blocks,
+                "bytes": dir_blocks * BLOCK_SIZE,
+            }
```
stacywsmith: We need to account for the possibility that the `<used-space>` element (or `used-blocks` attribute) doesn't exist. I suggest replacing lines 292-298 with:
```
used_space = directory.find('used-space')
if used_space is not None:
    dir_size = used_space.text()
    dir_blocks = used_space.get('used-blocks')
    if dir_blocks is not None:
        dir_bytes = dir_blocks * BLOCK_SIZE
        result[dir_name] = {
            "size": dir_size,
            "blocks": dir_blocks,
            "bytes": dir_bytes,
        }
```

lib/jnpr/junos/utils/fs.py
pklimai: @stacywsmith  using normalize=True was my initial intention, but then test test_directory_usage() fails with reason that is completely unclear to me:

```
Error
  File "/usr/local/lib/python3.5/dist-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/home/peter/PycharmProjects/py-junos-eznc/tests/unit/utils/test_fs.py", line 272, in test_directory_usage
    self.assertEqual(self.fs.directory_usage(path="/var/tmp"),
  File "/home/peter/PycharmProjects/py-junos-eznc/lib/jnpr/junos/utils/fs.py", line 281, in directory_usage
    rsp = self._dev.rpc.get_directory_usage_information(path=path, depth=str(depth), normalize=True)
  File "/home/peter/PycharmProjects/py-junos-eznc/lib/jnpr/junos/rpcmeta.py", line 156, in _exec_rpc
    return self._junos.execute(rpc, **dec_args)
  File "/usr/local/lib/python3.5/dist-packages/mock/mock.py", line 1062, in __call__
    return _mock_self._mock_call(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/mock/mock.py", line 1128, in _mock_call
    ret_val = effect(*args, **kwargs)
  File "/home/peter/PycharmProjects/py-junos-eznc/tests/unit/utils/test_fs.py", line 338, in _mock_manager
    device_params = kwargs['device_params']
KeyError: 'device_params'
```

I see other utility methods use strip() rather extensively - I guess this is probably due to the same reason.

lib/jnpr/junos/utils/fs.py
pklimai: @stacywsmith findall('//directory') fails with error,

```
  File "/usr/lib/python3/dist-packages/lxml/_elementpath.py", line 246, in _build_path_iterator
    raise SyntaxError("cannot use absolute path on element")
SyntaxError: cannot use absolute path on element
```

that's why I was using xpath() initially. However findall('.//directory') works, so I will use it.

pklimai:
Pushed latest version, please review. As I said before I wasn't able to use "normalize" option (possibly a separate issue?). Instead using text.strip(), hopefully this is rather safe.

Also note that in this version for non-existing "path" we just return {}, is it what we want?

vnitinv:
@pklimai Can you please fine tune your tests as its failing for Python 3.x
Thanks

pklimai:
@vnitinv That was some problem with the test system - test_sw_install_kwargs_force_host was failing, that I didn't touch. Now all tests passed.

vnitinv:
Thanks @pklimai.

PR #630, opened 2016-12-05
Issu sw install: pre-requisite requirements to be taken care by the function itself

vnitinv:
```python
with Device(host='xxxx', user='xxxx', password='xxxx') as dev:
    pprint(dev.facts)
    sw = SW(dev)
    try:
        ok = sw.install(package='junos-install-mx-x86-64-16.1-20160925.0.tgz', issu=True,
                  no_copy=True, validate=True, progress=True)
        print 'install Status %s' % ok
        if ok is True:
            # below step is only required you we want to reboot the old master re.
            sw.reboot(all_re=False)
            print 'master re rebooted'
            # device might be pingable for few seconds before it goes for reboot.
            time.sleep(30)
            if dev.probe(1200, 5):
                print "old master is rebooted and is now up"
                dev.open(gather_facts=False)
                time.sleep(300) # without sleep we get message "Not ready for mastership switch, try after 239 secs."
                # un-comment below line if we want to continue with initial master now also.
                # print dev.cli("request chassis routing-engine master acquire", warning=False)
                print "Device is UP & ISSU upgrade Done"
            else:
                print "old master is not pingable even after 1200 sec after reboot"
        else:
            "sw.install returned False"
    except Exception as ex:
        print "ERROR: %s" % ex.message
```

### DIFF
--- a/lib/jnpr/junos/utils/sw.py
+++ b/lib/jnpr/junos/utils/sw.py
@@ -264,43 +264,46 @@ class SW(Util):
                  'backup Routing engine must be running the same software\n'
                  'version before you can perform a unified ISSU.')
         if not (self._dev.facts['2RE'] and self._dev.facts['version_RE0'] == self._dev.facts['version_RE1']):
-            self.log('The master Routing Engine and backup Routing Engine must be running\n'
-                     'the same software version before it can perform a unified ISSU')
+            self.log('Requirement FAILED: The master Routing Engine (%s) and \n'
+                     'backup Routing Engine (%s) must be running the same \n'
+                     'software version before it can perform a unified ISSU' %
+                     (self._dev.facts['version_RE0'],
+                      self._dev.facts['version_RE1']))
             return False
+        self.log('Checking GRES status')
         conf = self._dev.rpc.get_config(filter_xml=etree.XML(
             '<configuration><chassis><redundancy><graceful-switchover/></redundancy></chassis></configuration>'))
-        self.log('Checking GRES status')
         if conf.find('chassis/redundancy/graceful-switchover') is None:
-            self.log('GRES is not Enabled in configuration')
+            self.log('Requirement FAILED: GRES is not Enabled in configuration')
             return False
         self.log('Checking NSR status')
         conf = self._dev.rpc.get_config(filter_xml=etree.XML(
             '<configuration><routing-options><nonstop-routing/></routing-options></configuration>'))
         if conf.find('routing-options/nonstop-routing') is None:
-            self.log('NSR is not Enabled in configuration')
+            self.log('Requirement FAILED: NSR is not Enabled in configuration')
             return False
         self.log('Checking commit synchronize status')
         conf = self._dev.rpc.get_config(
             filter_xml=etree.XML('<configuration><system><commit><synchronize/></commit></system></configuration>'))
         if conf.find('system/commit/synchronize') is None:
-            self.log('commit synchronize is not Enabled in configuration')
+            self.log('Requirement FAILED: commit synchronize is not Enabled in configuration')
             return False
-        self.log('Verifying that NSR is configured on the master Routing Engine\n'
-                 '(re0) by using the "show task replication" command.')
+        self.log('Verifying that NSR is configured on the current Routing Engine\n'
+                 'by using the "show task replication" command.')
         op = self._dev.rpc.get_routing_task_replication_state()
         if not (op.findtext('task-gres-state') == 'Enabled' and op.findtext('task-re-mode') == 'Master'):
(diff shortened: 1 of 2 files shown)


lib/jnpr/junos/utils/sw.py
```diff
@@ -238,6 +242,71 @@ def validate(self, remote_package, issu=False, **kwargs):
...
+        if not (self._dev.facts['2RE'] and self._dev.facts['version_RE0'] == self._dev.facts['version_RE1']):
+            self.log('Requirement FAILED: The master Routing Engine (%s) and \n'
+                     'backup Routing Engine (%s) must be running the same \n'
+                     'software version before it can perform a unified ISSU' %
+                     (self._dev.facts['version_RE0'],
+                      self._dev.facts['version_RE1']))
+            return False
+        self.log('Checking GRES status')
+        conf = self._dev.rpc.get_config(filter_xml=etree.XML(
+            '<configuration><chassis><redundancy><graceful-switchover/></redundancy></chassis></configuration>'))
```
stacywsmith: Any time we are checking the config, we need to be very careful. We must account for configuration groups, commit scripts, etc. We also must check the committed configuration rather than the default candidate configuration.

This change would handle committed config, config groups, and commit scripts:
```
    conf = self._dev.rpc.get_config(filter_xml=etree.XML(
        '<configuration><chassis><redundancy><graceful-switchover/></redundancy></chassis></configuration>')
        options={'database': 'committed', 'inherit': 'inherit', 'commit-scripts': 'apply',})
```

However, there are still multiple ways this could fail. This includes ephemeral DB and a lack of permission to view the configuration.

Whenever possible, we should try to check with operational commands rather than configuration checks. In this case, I don't see a way to avoid the configuration check, but we should at least consider there are known possibilities for GRES to be correctly configured, but this check to fail.

lib/jnpr/junos/utils/sw.py
```diff
@@ -238,6 +242,71 @@ def validate(self, remote_package, issu=False, **kwargs):
...
+        if conf.find('chassis/redundancy/graceful-switchover') is None:
+            self.log('Requirement FAILED: GRES is not Enabled in configuration')
+            return False
+        self.log('Checking NSR status')
+        conf = self._dev.rpc.get_config(filter_xml=etree.XML(
+            '<configuration><routing-options><nonstop-routing/></routing-options></configuration>'))
```
stacywsmith: See comment on line 275.

lib/jnpr/junos/utils/sw.py
```diff
@@ -238,6 +242,71 @@ def validate(self, remote_package, issu=False, **kwargs):
...
+        if conf.find('routing-options/nonstop-routing') is None:
+            self.log('Requirement FAILED: NSR is not Enabled in configuration')
+            return False
+        self.log('Checking commit synchronize status')
+        conf = self._dev.rpc.get_config(
+            filter_xml=etree.XML('<configuration><system><commit><synchronize/></commit></system></configuration>'))
```
stacywsmith: See comment on line 275.

lib/jnpr/junos/utils/sw.py
```diff
@@ -238,6 +242,71 @@ def validate(self, remote_package, issu=False, **kwargs):
...
+    def _issu_requirement_validation(self):
+        """
+        Checks:
+            * The master Routing Engine and backup Routing Engine must be
+                running the same software version before you can perform a
+                unified ISSU.
+            * Check GRES is enabled
+            * Check NSR is enabled
+            * Check commit synchronize is enabled
+            * Verify that NSR is configured on the master Routing Engine (re0)
```
stacywsmith: Remove "(re0)"

lib/jnpr/junos/utils/sw.py
```diff
@@ -238,6 +242,71 @@ def validate(self, remote_package, issu=False, **kwargs):
...
+                by using the "show task replication" command.
+            * Verify that GRES is enabled on the backup Routing Engine (re1)
```
stacywsmith: Remove "(re1)"

vnitinv: Done

lib/jnpr/junos/utils/sw.py
vnitinv: Done the changes as per your suggestion

lib/jnpr/junos/utils/sw.py
vnitinv: Done the changes as per your suggestion

lib/jnpr/junos/utils/sw.py
vnitinv: Done the changes as per your suggestion

PR #631, opened 2016-12-05
Improve error handling in directory_usage():

stacywsmith:
Improve error handling in directory_usage():
 - Raise an RpcError if no <directory> element present in the response.
 - Raise an RpcError (rather than trying to perform None.strip()) if a
   <directory-name> element is missing.


stacywsmith:
@vnitinv Can you please review? This is just a slight enhancement to #629.

PR #638, opened 2016-12-20
Enhanced fact gathering

stacywsmith:
- Enhanced the fact gathering system to perform on-demand fact gathering.
- Modified how some of the facts are gathered.
- Added some new facts.

The new-style on-demand fact gathering is the default. The old-style fact gathering can still be used (for now) by passing `fact_style='old'` to `Device()`.


stacywsmith:
Still need to add unit tests for new code.

stacywsmith:
> The routing engine keys are lowercase, although in all other places (e.g.: 'master': 'RE0' or 'RE0':  {'status': 'OK', [...]}) it's upper.
> Same under hostname_info and model_info.

This was actually intentional. For any existing facts, I maintained the current behavior (mostly capitalized) to ensure backwards compatibility. For new facts like `hostname_info` and `model_info` I maintained the case of the keys as they actually come from the device.

> the information for RE0 and RE1 is duplicated and appears as keys 0 and 1 represented as both int and str.

This was also intentional, but I'm now rethinking that decision. The idea was that it wouldn't matter if the user did: `dev.facts['re_info']['default'][0]['status']` or `dev.facts['re_info']['default']['0']['status']`. In other words, the slot could be specified as an integer or a string. However, the impact of this decision is that you can't loop over all of the slots without having to deal with the duplicates. I suspect I will remove the integer version and just keep the string version since all other keys are strings anyway. Your input on that decision is welcome.

> Also: it's not entirely clear to me what is supposed to be the default key at the root of re_info?

That's an attempt to deal consistently with both single and multi-chassis systems. The first-level key represents the chassis/node.  Since single chassis systems don't have a chassis/node name, I chose to represent that as `'default'`.

One of the things I also tried to do in this commit was accurately document each fact (both the already existing ones and the new ones I added). This documentation includes more detail on the keys for the new `hostname_info`, `model_info`, `junos_info`, and `current_re` facts I added. Once merged, the documentation for these facts will be part of the auto-generated "read the docs" documentation. Until then you can clone the repo, `cd docs`, `make html`, and open the resulting `<repo>/docs/_build/html/index.html` to view the documentation.

mirceaulinic:
> This was actually intentional. For any existing facts, I maintained the current behavior (mostly capitalized) to ensure backwards compatibility. For new facts like hostname_info and model_info I maintained the case of the keys as they actually come from the device.

Consistency should be intentional as well - at least IMHO.

> This was also intentional, but I'm now rethinking that decision. The idea was that it wouldn't matter if the user did: dev.facts['re_info']['default'][0]['status'] or dev.facts['re_info']['default']['0']['status'].

```int``` type keys in a python dict are at the opposite side of best practices. Also transforming the object into JSON would overwrite the keys anyways, so it doesn't make much sense to me.

> That's an attempt to deal consistently with both single and multi-chassis systems. The first-level key represents the chassis/node. Since single chassis systems don't have a chassis/node name, I chose to represent that as 'default'.

Thanks!

>  Once merged, the documentation for these facts will be part of the auto-generated "read the docs" documentation. Until then you can clone the repo, cd docs, make html, and open the resulting <repo>/docs/_build/html/index.html to view the documentation.

Yes, I will do that next week and I will let you know again if I notice anything suspicious.

Cheers,
Mircea

**Comment by vnitinv** [comment]

File: lib/jnpr/junos/device.py [hunk 1]
```diff
@@ -638,27 +641,72 @@ def execute(self, rpc_cmd, **kvargs):
...
+                                         errors out.
+                                         warnings_on_failure=True is the
+                                         default for old-style facts.
+                                         warnings_on_failure=False is the
+                                         default for new-style facts.
+                                         If True when new-style fact gathering
+                                         is in use, causes all facts to be
+                                         reloaded rather than being loaded on
+                                         demand.
+        :param str, set, list, or tuple keys: The set of keys in facts to
```
vnitinv: Can we add an example for this option?
Any why we have this rule, "and exception_on_failure==False and warnings==False"

lib/jnpr/junos/device.py
```diff
@@ -638,27 +641,72 @@ def execute(self, rpc_cmd, **kvargs):
...
-            except:
-                if exception_on_failure:
-                    raise
-                should_warn = True
-        if should_warn is True:
-            warnings.warn('Facts gathering is incomplete. '
-                          'To know the reason call '
-                          '"dev.facts_refresh(exception_on_failure=True)"',
-                          RuntimeWarning)
+        if (self._fact_style != 'old' and
```
vnitinv: We can just do 

if self._fact_style not in ['old', 'new', 'both']:

lib/jnpr/junos/factcache.py
```diff
@@ -0,0 +1,280 @@
...
+      * :meth:`__setitem__`: Called when a key is set on the dict.
+      * :meth:`__iter__`: Called when iterating over the keys of the dict.
+      * :meth:`__len__`: Called when getting the length of the dict.
+      * :meth:`__repr__`: Called when representing the dict as a string.
+
+    **Additional methods:**
+      * :meth:`_refresh`: Refreshes the fact cache.
+    """
+    # Used to register the class as a dict.
+    __metaclass__ = ABCMeta
```
vnitinv: Have you tested this on Python 3.x. I remember there were some syntax changes to __metaclass__

vnitinv: check this link
http://python-future.org/compatible_idioms.html#metaclasses

lib/jnpr/junos/factcache.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        Facts are read-only. Don't allow deleting an item.
+        """
+        raise RuntimeError("facts are read-only!")
+
+    def __setitem__(self, key, value):
+        """
+        Facts are read-only. Don't allow setting an item.
+        """
+        raise RuntimeError("facts are read-only!")
```
vnitinv: In older version we can do

dev.facts['serialnumber'] = 'xxxx'
This wont throw any exception and had been used by many user (Eddie) as vMX dont used to have serial number

dev.facts = 'xxxx' this used to throw exception. 

but with current changes even dev.facts['serialnumber'] = 'xxxx' will throw exception.
Are we sure of keeping this way?

lib/jnpr/junos/factcache.py
```diff
@@ -0,0 +1,280 @@
+from abc import ABCMeta
+import collections
+import warnings
+
+import jnpr.junos.facts
+import jnpr.junos.exception
+
+
+class _FactCache(collections.MutableMapping):
```
vnitinv: Now we get

type(dev.facts)
<class 'jnpr.junos.factcache._FactCache'>

Hope user is not comparing it with dict anywhere

lib/jnpr/junos/factcache.py
mirceaulinic: +1

I also vote to keep the current behaviours unchanged.
Also my previous comment about keeping the uppercase/lowercase consistent still stands IMO.

lib/jnpr/junos/facts/__init__.py
```diff
@@ -1,24 +1,108 @@
...
+
+    :returns:
+      A list of fact module names.
+    """
+    module_names = []
+    facts_dir = os.path.dirname(__file__)
+    for file in os.listdir(facts_dir):
+        if (os.path.isfile(os.path.join(facts_dir, file)) and
+           not os.path.islink(os.path.join(facts_dir, file))):
+            if file.endswith('.py') and not file.startswith('_'):
```
vnitinv: glob module was better option. But thats fine
>>> import glob
>>> glob.glob('*.py')

stacywsmith: I did consider using `glob.glob()`, but I ultimately chose this method because there's no single glob pattern which would satisfy `file.endswith('.py') and not file.startswith('_')`. (At least I'm not aware of a way to specify negation in a glob pattern.) 

In addition, `glob.glob('*.py')` returns the .py files in the current directory rather than in `facts_dir`. 

If I do:
```
for file in glob.glob(facts_dir + '/*.py'):
```
Then the value of file:
1) contains the path which has to be stripped off in an extra step.
2) still has to be checked to make sure it's not a directory and not a link.
3) still has to be checked to make sure it doesn't start with `'_'`.

lib/jnpr/junos/factcache.py
```diff
@@ -0,0 +1,280 @@
...
+                                  'To know the reason call '
+                                  '"dev.facts_refresh('
+                                  'exception_on_failure=True)"',
+                                  RuntimeWarning)
+                self._exception_on_failure = False
+                self._warnings_on_failure = False
+                self._should_warn = False
+
+# Make this class look like a regular dict.
+# Ensures isinstance() and issubclass() behave the same as for a dict.
```
vnitinv: type(dev.facts)
<class 'jnpr.junos.factcache._FactCache'>

isinstance(dev.facts, dict) return False

lib/jnpr/junos/device.py
stacywsmith: Addressed. Added examples to the documentation, and documented the reason for the restriction. Either `exception_on_failure` or `warnings_on_failure` imply that ALL facts will be refreshed so it doesn't make sense to specify specific keys in those cases.

lib/jnpr/junos/device.py
stacywsmith: Addressed.

lib/jnpr/junos/factcache.py
stacywsmith: You are correct that this was not working as expected. The intention was to make the `_FactCache` class appear as a subclass of `dict`, but this was not working as expected and has been removed.

lib/jnpr/junos/factcache.py
stacywsmith: I have not yet changed this behavior. Since these are supposed to be "facts" about the device it seems inappropriate to allow users to set a fact to an arbitrary value. The `facts_refresh()` method already provides a way to force the refresh of specific facts.

Can you please explain the use case for setting "facts" to a user-defined value?

lib/jnpr/junos/factcache.py
stacywsmith: Yes. This is what I was trying to work around with metaclasses. However, that did not work as expected. I'm open to any suggestions on how to workaround this restriction. However, since direct type-testing is not recommended best practice in Python, and this class does support duck typing, I don't expect this to be a significant issue.

lib/jnpr/junos/factcache.py
stacywsmith: Yes. This is what I was trying to workaround with using the ABCMeta metaclass. Unfortunately, what I had implemented was the inverse of what was intended (an instance of a dict was an instance of _FactCache, but not vice-versa.)

PR #645, opened 2017-01-09
added remote port ID to lldp.yml (OP)

Jaredn:
Added Remote Port ID to lldp.yml OP.  Not sure why it wasn't there to begin with.  Tested against live MX480.


stacywsmith:
Thanks @Jaredn

Jaredn:
Thanks for the fast merge!

PR #647, opened 2017-01-12
Fix documentation for rollback()

vincentbernat:
The expected rollback ID is an integer.


stacywsmith:
Thanks @vincentbernat.

PR #648, opened 2017-01-13
Do not provide the key if it comes from the SSH configuration

vincentbernat:
This is a followup of #628 where this change was initially pushed but
reverted because I didn't remember why I did it. Before #628, when an
identity is provided in the SSH configuration, it was not copied in
`_conf_ssh_private_key_file` due to a bug. After fixing the bug in #628,
the key is now copied.

However, the SSH configuration is provided to the `connect()` method
which will use it if needed. Therefore, this is not needed. Moreover, if
the key is provided by an agent and/or encrypted, this won't work as,
later in the code, `allow_agent` will be set to `False` due to the
presence of a private key.

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -218,7 +218,6 @@ class _Connection(object):
                 self._hostname = found.get('hostname', self._hostname)
                 self._port = found.get('port', self._port)
                 self._conf_auth_user = found.get('user')
-                self._conf_ssh_private_key_file = found.get('identityfile')
             return sshconf_path
 
     def display_xml_rpc(self, command, format='xml'):
@@ -830,15 +829,13 @@ class Device(_Connection):
             # user will default to $USER
             self._auth_user = os.getenv('USER')
             self._conf_auth_user = None
-            self._conf_ssh_private_key_file = None
             # user can get updated by ssh_config
             self._ssh_config = kvargs.get('ssh_config')
             self._sshconf_lkup()
             # but if user or private key is explicit from call, then use it.
             self._auth_user = kvargs.get('user') or self._conf_auth_user or \
                 self._auth_user
-            self._ssh_private_key_file = kvargs.get('ssh_private_key_file') \
-                or self._conf_ssh_private_key_file
+            self._ssh_private_key_file = kvargs.get('ssh_private_key_file')
             self._auth_password = kvargs.get(
                 'password') or kvargs.get('passwd')
 


vincentbernat:
An alternative solution would be to enable the user to say if they want to use the agent or not. `allow_agent=None` in the function signature. `if allow_agent is None`, use the current heuristic to decide its value, otherwise leave it as is.

IMO, we should always use the agent: `ssh` doesn't offer the option to not query the agent but I suppose people put that for some reason.

jnpr-community-netdev:

vnitinv:

lib/jnpr/junos/device.py
```diff
@@ -218,7 +218,6 @@ def _sshconf_lkup(self):
                 self._hostname = found.get('hostname', self._hostname)
                 self._port = found.get('port', self._port)
                 self._conf_auth_user = found.get('user')
-                self._conf_ssh_private_key_file = found.get('identityfile')
```
vnitinv: Not sure why is this lined removed. 
So if I have entered my private ssh file as *identityfile* in my ssh cofig. witout this line those private key file will be ignored and its like everytime user need to pass with ssh_private_key_file parameter. 

In case of multiple device usage with different private file for different device. identityfile is of much use.

vincentbernat: The information is still present in the SSH configuration that is provided to ncclient. So, your use case still work.

The line is removed because when `_conf_ssh_private_key_file` is defined, it disables the SSH agent (which is quite surprising as no other SSH client does that). So, if the key is stored in the agent, that doesn't work.

vnitinv:

vincentbernat:
Is it something to do for this PR to be merged? People using an agent to store their SSH keys cannot use junos-eznc without this patch as agent is bypassed when providing a key directly to junos-eznc or through `~/.ssh/config`.

vnitinv:
ok to test

jnpr-community-netdev:
Build triggered. sha1 is merged.

jnpr-community-netdev:
Build started sha1 is merged.

jnpr-community-netdev:
Build finished. 730 tests run, 0 skipped, 0 failed.

jnpr-community-netdev:
Build finished.

vnitinv:
Sorry for the delay @vincentbernat I need to test this change before I can merge. I will do it next week.

vincentbernat:
This bug is still biting me 2 years later. Again, the root cause of all this pain is the ability to disable querying local agent. This is not an option the regular `ssh` client allows. An easy fix would be to always set to `True` and see who complains.

vnitinv:
Sorry @vincentbernat , we will update you about this by EOD today with the code changes we plan. You can go through the plan. If all good, will take it forward.

rsmekala:
@vincentbernat I am sorry for the huge delay. We will be merging the required changes this week.

Regarding the logic, we have decided to go with `allow_agent` logic as suggested in the discussion. I have validated the changes for a generic setup. Will validate the changes for console server setup.

rsmekala:
@vincentbernat @paravoid The functionality will be merged via #920.  Can you please review the changes, and let us know if you have any additional requirements ??

gaima8:
https://github.com/Juniper/py-junos-eznc/pull/1284 is another go at resolving this.
In my case I have an IdentityFile defined at the top level of my ~/.ssh/config but for the Juniper devices I must use a yubikey so have no private key on disk to use for them.

With `allow_agent` defined it simply won't look for an IdentityFile and pass allow_agent=True to ncclient.

vincentbernat:
I wanted to get consensus here before opening yet another pull request, but after years, still no consensus. Nobody was able to explain why we try to disable the agent, something that is not possible with "ssh" from OpenSSH. I think the best way forward is to mimic this and always leave the agent enabled. Nobody ever demonstrated this would break their workflow. Adding more options and workarounds will just makes the code more complex and brittle. I have opened #1285 to remove the logic around disabling the agent and it should be favored over of the current PR.

PR #649, opened 2017-01-13
Make SCP and StartShell behaves the same way as Device for SSH

vincentbernat:
This is a followup to #648 (and this PR includes the commit from #648, so it shouldn't be merged before #648).

First commit makes small changes to SCP. This is mostly cosmetic with the exception of the agent and the private key. This is quite similar to what ncclient is doing with the same parameters that were provided in `Device`.

The second commit moves the creation of the paramiko SSH client into a dedicated function.

The third commit reuses this dedicated function for `StartShell()`.

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -819,6 +819,7 @@ class Device(_Connection):
             self._hostname = 'localhost'
             self._ssh_private_key_file = None
             self._ssh_config = None
+            self._allow_agent = False
         else:
             # --------------------------
             # making a remote connection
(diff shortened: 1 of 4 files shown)


jnpr-community-netdev:

vnitinv:

vnitinv:
ok to test.

lib/jnpr/junos/device.py
```diff
@@ -830,17 +830,21 @@ def __init__(self, *vargs, **kvargs):
...
                 self._auth_user
-            self._ssh_private_key_file = kvargs.get('ssh_private_key_file') \
-                or self._conf_ssh_private_key_file
+            self._ssh_private_key_file = kvargs.get('ssh_private_key_file')
             self._auth_password = kvargs.get(
                 'password') or kvargs.get('passwd')
+            # we want to enable the ssh-agent if-and-only-if we are
+            # not given a password or an ssh key file.
+            # in this condition it means we want to query the agent
+            # for available ssh keys
```
vincentbernat: If you provide a password, the agent is disabled and Paramiko will fetch the key and decrypt it. The agent doesn't allow you to provide a password for an encrypted key either (it will ask the user for that if needed). Moreover, this code is already present as is. It's just moved here to be able to reuse its result for scp.

IMO, the ability to disable the agent part is odd. The regular "ssh" client doesn't have such an option. It adds complexity and I don't know why this has been implemented. We should just get rid of that.

vnitinv:

vnitinv:
ok to test

vnitinv:
@stacywsmith Can you please review this pull request.

jnpr-community-netdev:
Build finished. 551 tests run, 0 skipped, 0 failed.

PR #658, opened 2017-02-03
fix Device.timeout is ignored when dev.open()

ohsugan:
In method comment, connection timeout value should be set by Device.timeout.
But, It's ignored.

>         :raises ConnectTimeoutError:
>             When the the :meth:`Device.timeout` value is exceeded
>             during the attempt to connect to the remote device

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -997,8 +997,8 @@ class Device(_Connection):
                                (self._ssh_private_key_file is None))
 
             # set connection timeout by Device.timeout                         
-            timeout = (self.__class__.timeout) \                                
-                    if isinstance(self.__class__.timeout, int) else (None)      
+            timeout = (self.__class__.timeout) \
+                    if isinstance(self.__class__.timeout, int) else (None)
 
             # open connection using ncclient transport
             self._conn = netconf_ssh.connect(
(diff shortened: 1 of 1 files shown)


jnpr-community-netdev:

vnitinv:
@ohsugan as per ncclient documentation 
> *timeout* is an optional timeout for socket connect

which internally call sock.settimeout(timeout)
Right now sock is non blocking for us. But if someone passes 1 as timeout then it becomes blocking call.
https://docs.python.org/2/library/socket.html#socket.socket.setblocking

We have defined timeout parameter for rpc timeout. 
Can I ask, What is the scenario in which you felt you need to set socket timeout.

ohsugan:
@vnitinv thanks to comment!
I'd like to collect configurations from many routes, which I don't know whether they are online or not.
So, I'd like to set different value on connection timeout and rpc timeout.

In this case, Is it better to handle connection timeout in application script?

vnitinv:
Use probe functionality for such cases. 
http://junos-pyez.readthedocs.io/en/2.0.0/jnpr.junos.html#jnpr.junos.device.Device.auto_probe

ohsugan:
thanks @vnitinv , I see.
According to Device#open() comment, connection timeout depends on Device.timeout(),
> ConnectTimeoutError – When the the Device.timeout() value is exceeded during the attempt to connect to the remote device

but, actually it depends on other parameter doesn't it?

Issue #660, opened 2017-02-06
with latest fact changes, pprint is not printing facts in pretty print manner


vnitinv:
```python
from pprint import pprint

with Device('xx.xx.xx.xx', user='xxxx') as dev:
    pprint(dev.facts)
```

{'2RE': True, 'HOME': '/var/home/regress', 'RE0': {'status': 'OK', 'last_reboot_reason': 'Router rebooted after a normal shutdown.', 'model': 'RE-VMX', 'up_time': '37 minutes, 8 seconds', 'mastership_state': 'master'}, 'RE1': {'status': 'OK', 'last_reboot_reason': 'Router rebooted after a .......

stacywsmith:
Fixed via #661.

Issue #651, opened 2017-01-19
PyEZ doesn't correctly process release format: 15.1F5.15-C1.12


ydnath:
PyEZ doesn't like the new naming convention in 15.1F5.15-C1.12:
The error I saw is 'ValueError: invalid literal for int() with base 10: '1F5''

In [1]: from jnpr.junos import Device

In [2]: dev = Device(host='sjcx1-bbr1')

In [3]: dev.open()
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/home/bocao/<ipython-input-3-ccff1d66f9df> in <module>()
----> 1 dev.open()

/usr/lib/python2.7/dist-packages/jnpr/junos/device.pyc in open(self, *vargs, **kvargs)
    493         gather_facts = kvargs.get('gather_facts', self._gather_facts)
    494         if gather_facts is True:
--> 495             self.facts_refresh()
    496
    497         return self

/usr/lib/python2.7/dist-packages/jnpr/junos/device.pyc in facts_refresh(self)
    771         """
    772         for gather in FACT_LIST:
--> 773             gather(self, self._facts)
    774
    775     # ------------------------------------------------------------------------

/usr/lib/python2.7/dist-packages/jnpr/junos/facts/swver.pyc in facts_software_version(junos, facts)
    193     # ------------------------------------------------------------------------

    194
--> 195     facts['version_info'] = version_info(facts['version'])
    196
    197

/usr/lib/python2.7/dist-packages/jnpr/junos/facts/swver.pyc in __init__(self, verstr)
      9         self.type = m1.group(2)
     10
---> 11         self.major = tuple(map(int, m1.group(1).split('.')))  # creates tuyple
     12         after_type = m1.group(3).split('.')
     13         self.minor = after_type[0]

ValueError: invalid literal for int() with base 10: '1F5'

vnitinv:
@stacywsmith With our latest changes for facts gathering, are we handling these cases too?

stacywsmith:
@vnitinv The current code does improve the handling of this, but I think we should continue to improve it further.

The current code correctly provides the version string, including the `C1.12`, as text. However, it does not parse that portion of the version string text into a `version_info` object. It simply omits the `C1.12` portion from the `version_info` object.

Here's the output of `dev.facts['junos_info']` from a router which is running 15.1F5.15-C1.12 on RE0 and 15.1F5.15 on RE1.

```
    're0': {
[...]

PR #661, opened 2017-02-06
fix for pprint #660

vnitinv:
Now print print and pprint should work as expected.

### DIFF
--- a/lib/jnpr/junos/factcache.py
+++ b/lib/jnpr/junos/factcache.py
@@ -221,12 +221,7 @@ class _FactCache(collections.MutableMapping):
           side-effect of causing any ungathered facts to be gathered and then
           cached.
         """
-        # below 3 line will help in populating _cache. Otherwise pprint
-        # sometime return empty dictionary
-        for key in sorted(self):
-            if not key.startswith('_'):
-                self.get(key)
-        return pformat(self._cache)
+        return pformat(dict(self))
 
     def _refresh(self,
                  exception_on_failure=False,
--- a/lib/jnpr/junos/factcache.py
+++ b/lib/jnpr/junos/factcache.py
@@ -1,5 +1,6 @@
 import collections
 import warnings
+from pprint import pformat
 
 import jnpr.junos.facts
 from jnpr.junos.facts import __doc__ as facts_doc
@@ -192,7 +193,7 @@ class _FactCache(collections.MutableMapping):
         """
         return len(self._callbacks)
 
-    def __repr__(self):
+    def __str__(self):
         """
         A string representation of the facts dictionary.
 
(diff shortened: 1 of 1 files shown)


lib/jnpr/junos/factcache.py
```diff
@@ -192,7 +193,7 @@ def __len__(self):
         """
         return len(self._callbacks)
 
-    def __repr__(self):
+    def __str__(self):
```
stacywsmith: Based on these definitions, I believe these two function definitions should be reversed.
`__str__()` should be the formatted and nicely printable representation.
https://docs.python.org/2/library/functions.html#str
`__repr__()` should be the string which could be passed back to `eval()`
https://docs.python.org/2/library/functions.html#repr

lib/jnpr/junos/factcache.py
```diff
@@ -211,6 +212,22 @@ def __repr__(self):
...
+          Because this returns the value of every fact, it has the
+          side-effect of causing any ungathered facts to be gathered and then
+          cached.
+        """
+        # below 3 line will help in populating _cache. Otherwise pprint
+        # sometime return empty dictionary
+        for key in sorted(self):
+            if not key.startswith('_'):
+                self.get(key)
+        return pformat(self._cache)
```
stacywsmith: Your implementation causes the "internal" facts (the facts whose names begin with an underscore) to also be printed. In addition, I believe the whole method implementation can be reduced to:
```
return pformat(dict(self))
```

lib/jnpr/junos/factcache.py
vnitinv: "Based on these definitions, I believe these two function definitions should be reversed.
__str__() should be the formatted and nicely printable representation."

I kept it normal as earlier too facts used to be printed in a line only

"__repr__() should be the string which could be passed back to eval()
https://docs.python.org/2/library/functions.html#repr"
The kind of facts we have(containing objects inside dict), that cant be EVALuated.
For me it use to throw SyntexError when used eval with fats sring.

lib/jnpr/junos/factcache.py
vnitinv: Thanks Stacy, This is even what I wanted.

lib/jnpr/junos/factcache.py
stacywsmith: Yes, the string from `repr()` can not yet be passed to `eval()` successfully, but that's because of the string representation of the `jnpr.junos.version_info` objects. We can fix that when I address #651. If we keep the function the `__repr__()` function as it was, and add your new function as `__str__()`, then we most closely mimic the previous behavior when `dev.facts` was a true `dict`.

PR #662, opened 2017-02-07
adjust ConnectTimeoutError comment on Device#open() to implement

ohsugan:
According to the comment on Device#open(), ConnectTimeoutError raises when Device.timeout() value is exceeded.
But, it actually depends on OS determined timeout value, such as net.ipv4.tcp_syn_retries.


jnpr-community-netdev:

vnitinv:

rahkumar651991:
This is correct as per the document. Device.timeout is set and it is used for the timeout. It is not specific to OS environment timeout value. 
Closing this.

Issue #663, opened 2017-02-09
Feature request: keepalive or auto-reconnect


spidercensus:
I'm working to implement Salt proxy minions to Juniper devices in a customer network where idle-timeouts are configured for 5 minutes.  This means that if we don't send something into the session before the 5 minutes are up, the session dies and a new proxy minion must be built.

I have tried adding the keepalive directive to my ssh_config file, but this had no effect. 
```
Host * 
    ServerAliveInterval  
```

Based on the above, it seems like transport layer keepalive is not going to solve this problem.  The Netconf spec currently does not contain a keepalive operation, and the IETF mailing list seems to have agreed not to create one: `https://www.ietf.org/mail-archive/web/netconf/current/msg08888.html`

I see three options for solving this problem
1. No changes to the Netconf repo.  Keepalives implemented entirely in application.
2. Implement keepalive Device parameter and keepalive thread which executes some RPC on a set interval.  I don't like this option because it will fill up device logs and induce CPU churn
3. Add an auto_reconnect parameter to Device.  This would allow for a check to be performed before every RPC is executed to make sure that the underlying SSH transport is still connected and functioning.  If it is not, then call open() to get the transport up before running the RPC.  

I think the third option is the easiest to implement.  Would you consider a merge if I created this?

vnitinv:
@spidercensus I think such check (and consecutive action) should be taken care by user's code.
right now dev.connected value is static, I am planning to make to property hence the value will be returning the current state of connection. Using this value user can take action as per there need.

spidercensus:
I agree that Device.connected needs to become a property.  After my discussion with stacy, I was going to raise another request for that.  

There is value in building this feature into the execute() and cli() methods on demand.  Frameworks such as salt will be forced to check the connection state before every RPC request, which leads to greater overhead than an internal check only when the flag is turned on.

spidercensus:
`https://github.com/Juniper/py-junos-eznc/pull/664`

spidercensus:
This is implemented in pull #669

PR #664, opened 2017-02-09
Connected property

spidercensus:
This partly addresses [issue #663](https://github.com/Juniper/py-junos-eznc/issues/663)


spidercensus:
This converts the `Device.connected` attribute into a property that depends on the new `_connected` member value.  There is a getter and a setter defined.

There is also now a Session Listener attached to the session object in the netconf transport, which responds to errors in the Transport in the background by setting the `Device.connected` value to False and raising a `TransportError` exception. 

I also added a step in the `Device.cli` method to properly catch the ConnectClosedError, which before was just being blackholed into the `except Exception as ex:` and reported back as "invalid command".

lib/jnpr/junos/device.py
```diff
@@ -940,6 +966,14 @@ def __init__(self, *vargs, **kvargs):
     # -----------------------------------------------------------------------
     # Basic device methods
     # -----------------------------------------------------------------------
+    @property
+    def connected(self):
+        return self._connected
```
vnitinv: where is _connected initialized?

stacywsmith: @vnitinv line 959 in `__init__()` initializes `self._connected` to `False` because it's calling the `connected` setter.

lib/jnpr/junos/device.py
```diff
@@ -730,6 +733,30 @@ def facts_refresh(self,
...
+        pass
+
+    def errback(self, ex):
+        """Called when an error occurs.
+        Set the device's connected status to False, and then raise a TransportError.
+
+        :type ex: :exc:`Exception`
+        """
+        self._device.connected = False
+        raise NcErrors.TransportError
```
stacywsmith: I don't believe we should raise an exception here. It's not actually doing anything anyway because ncclient catches the exception and just attempts to log a debug message:
https://github.com/ncclient/ncclient/blob/master/ncclient/transport/session.py#L77-L80

A `TransportError` exception will already be raised the next time there's an attempt to send on this connection:
https://github.com/ncclient/ncclient/blob/master/ncclient/transport/session.py#L155

lib/jnpr/junos/device.py
```diff
@@ -730,6 +733,30 @@ def facts_refresh(self,
...
 
+class DeviceSessionListener(SessionListener):
+
+    """
+    Listens to Session class of Netconf Transport and detects errors in the transport.
+    """
+    def __init__(self, device):
+        self._device = device
+    def callback(self, root, raw):
+        """Called when a new XML document is received. The *root* argument allows the callback to determine whether it wants to further process the document.
```
stacywsmith: Please run through `pep8` and clean up any reported problems like this long line.

vnitinv:
@stacywsmith Give me 1-2 days to test this pull request. Then we can merge this.

lib/jnpr/junos/device.py
spidercensus: I think you're right.  I'll remove that.

PR #665, opened 2017-02-11
update ospf view, add bgp/inventory

itdependsnetworks:
Myself and @GGabriele put these together. It would be nice to have included in official repo so I don't have to set specific directory when getting these views.


jnpr-community-netdev:

vnitinv:
ok to test

PR #668, opened 2017-02-14
Implement SSH keepalives, default 30-second interval, for NETCONF over SSH sessions.

stacywsmith:
Without SSH keepalives, a NAT or stateful firewall along the network
path between the PyEZ host and the target Junos device,
may timeout an inactive TCP flow and cause the NETCONF over SSH
session to hang. Sending SSH keepalives avoids this situation. The
default value is 30 seconds. Setting this parameter to a value of 0
disables SSH keepalives.

Note: This is a different situation than Issue #663 in which the
target Junos device is timing out the NETCONF over SSH session
due to a configured idle-timeout on the system login class.

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -866,6 +866,16 @@ class Device(_Connection):
             This can be used to load SSH information from a configuration file.
             By default ~/.ssh/config is queried.
 
+        :param int ssh_keepalives:
+            *OPTIONAL* The interval, in seconds, at which to send SSH
+            keepalives for the NETCONF over SSH session. Without SSH
+            keepalives, a NAT or stateful firewall along the network
+            path between the PyEZ host and the target Junos device,
+            may timeout an inactive TCP flow and cause the NETCONF over SSH
+            session to hang. Sending SSH keepalives avoids this situation. The
+            default value is 30 seconds. Setting this parameter to a value of 0
+            disables SSH keepalives.
+
         :param bool normalize:
             *OPTIONAL* default is ``False``.  If ``True`` then the
             XML returned by :meth:`execute` will have whitespace normalized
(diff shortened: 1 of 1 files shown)


spidercensus:
There should be a check to make sure that this value in the ssh config file is not overridden.

lib/jnpr/junos/device.py
```diff
@@ -1008,6 +1020,9 @@ def open(self, *vargs, **kvargs):
                 ssh_config=self._sshconf_lkup(),
                 device_params={'name': 'junos', 'local': False})
 
+            self._conn._session.transport.set_keepalive(
```
vnitinv: line 910 we are setting it to None for on-box if condition. So on on-box are not reaching line 924, hence converting None to int will throw exception.

vnitinv:

lib/jnpr/junos/device.py
```diff
@@ -897,6 +907,7 @@ def __init__(self, *vargs, **kvargs):
             self._hostname = 'localhost'
             self._ssh_private_key_file = None
             self._ssh_config = None
+            self._ssh_keepalives = None
```
spidercensus: Setting this to ``None`` does not make sense.  This value is initialized to zero in the underlying module. 

Help on method set_keepalive in module paramiko.transport:

set_keepalive(self, interval) method of paramiko.transport.Transport instance
    Turn on/off keepalive packets (default is off).  If this is set, after
    ``interval`` seconds without sending any data over the connection, a
    "keepalive" packet will be sent (and ignored by the remote host).  This
    can be useful to keep connections alive over a NAT, for example.
    
    :param int interval:
        seconds to wait before sending a keepalive packet (or
        0 to disable keepalives).

vnitinv: None is set for on-box python, where it won't use ssh connection. So this should be OK

PR #669, opened 2017-02-15
Auto reconnect

spidercensus:
This branch adds autoreconnect to the Device class. It is built on the connected_property branch.  

An optional integer parameter `autoreconnect` is added to `Device.__init__`.  This value is handled by a new property getter and setter of the same name. 

This value is used in the `execute` method.  At the beginning of this method, it checks if `connected` is True.  Previously it would raise an Exception if it were not true.  

With this branch, this method checks if `autoreconnect` is greater than 0, and if so, it attempts to run a `self.open()`, decrements `autoreconnect`, and then runs recursively runs `self.execute` with the same parameters.  If it returns a result without throwing an exception, then `autoreconnect` is incremented, and the result is returned upwards.


spidercensus:
This implements `https://github.com/Juniper/py-junos-eznc/issues/663`

spidercensus:
I would be open to changing this to a boolean parameter instead of an int.  I'm not sure why we'd ever need to attempt the reconnect more than once... but I figured I'd build it that way just in case.  
@stacywsmith let me know what you think

lib/jnpr/junos/device.py
```diff
@@ -550,7 +553,14 @@ def execute(self, rpc_cmd, **kvargs):
         """
 
         if self.connected is not True:
-            raise EzErrors.ConnectClosedError(self)
+            if self.autoreconnect > 0:
+                self.autoreconnect -= 1
+                self.open()
```
stacywsmith: If the reconnect fails, this call will raise one of the exceptions from `open()`. These exceptions are not caught, effectively making any value for `autoreconnect > 1` useless. In addition, the caller of `execute()` is not expecting these new exception types to be raised.

Instead, I believe lines 557 and 558 should be changed to something like:
```
try:
    self.open()
# All exceptions returned by open() are a subclass of ConnectError
except EzErrors.ConnectError:
    self.autoreconnect -= 1
```

lib/jnpr/junos/device.py
```diff
@@ -550,7 +553,14 @@ def execute(self, rpc_cmd, **kvargs):
         """
 
         if self.connected is not True:
-            raise EzErrors.ConnectClosedError(self)
+            if self.autoreconnect > 0:
```
stacywsmith: I don't think we should directly modify the `autoreconnect` value in this block. If `execute()` raises some exception such as a `RpcError` and the caller catches the exception, then we can end up in a situation where the value of `autoreconnect` get decremented, but never incremented.

stacywsmith:
> I would be open to changing this to a boolean parameter instead of an int. I'm not sure why we'd ever need to attempt the reconnect more than once... but I figured I'd build it that way just in case.
> @stacywsmith let me know what you think

I think I prefer the integer value. If the connection is closed by the target device due to inactivity, then a boolean would probably be sufficient. However, the connection might have failed due to an intermittent network problem. In that case, one reconnect attempt might not be sufficient and I'd prefer to let the user decide how many reconnect attempts is appropriate for their environment.

lib/jnpr/junos/device.py
```diff
@@ -940,6 +976,22 @@ def __init__(self, *vargs, **kvargs):
...
+        if value in [True, False]:
+            self._connected = value
+
+    @property
+    def autoreconnect(self):
+        return self._autoreconnect
+
+    @autoreconnect.setter
+    def autoreconnect(self, value):
+        self._autoreconnect = value
```
spidercensus: good idea.

lib/jnpr/junos/device.py
```diff
@@ -940,6 +976,22 @@ def __init__(self, *vargs, **kvargs):
...
     # Basic device methods
     # -----------------------------------------------------------------------
+    @property
+    def connected(self):
+        return self._connected
+
+    @connected.setter
+    def connected(self, value):
+        if value in [True, False]:
+            self._connected = value
```
spidercensus: 1.  I have the property setter there specifically to do type enforcement.  Otherwise, this would have been a member.  
2.  Yes, I suppose you're right.  Let me make some changes for htat.

lib/jnpr/junos/device.py
spidercensus: Yeah, this makes sense.  I think it might actually be better to catch any exceptions generated by the execute method and use the catch block to attempt a reconnect.  I'll get to work on that.

spidercensus:
The problem from CI was that the device was failing this test:
```    
def test_device_execute_unopened(self):
        self.dev.connected = False
        self.assertRaises(EzErrors.ConnectClosedError, self.dev.execute, None)```

If the device connection is not open and we attempt to run `execute` from a device which has autoreconnect reconfigured, then the device connection is automatically opened. That may not be desirable behavior.

mirceaulinic:
If that's required only in Salt, you may need to know that proxy keepalive will be available in Nitrogen (scheduled to be released in may) - see https://github.com/saltstack/salt/pull/38829

To see how to use this, you can have a look at: https://docs.saltstack.com/en/develop/topics/proxyminion/index.html - all you need to do is adding the `alive` function in the proxy module which returns a flag reflecting the state of the underlying transport (NETCONF in your case).

stacywsmith:
Thanks for the info on Salt. However, I feel this functionality is also useful in other PyEZ use cases.

vnitinv:

spidercensus:
If that is in fact implemented in Salt, then the new connect property from #664 should allow Salt to detect a drop and initiate the reconnection on its own.

rahkumar651991:
https://github.com/saltstack/salt/blob/master/salt/proxy/junos.py#L221 --> this takes care of the reconnect.

closing this issue based on this.

PR #681, opened 2017-03-15
Config.load update parameter option

vnitinv:
```python
    cu = Config(dev)
    cu.load(path='/var/tmp/xxx.conf', update=True)
```


vnitinv:
will also need to add this option to ansible galaxy module 'junos_install_config' as per request 
https://github.com/Juniper/ansible-junos-stdlib/issues/207

PR #682, opened 2017-03-16
Provide a master property and a re_name property for Device

stacywsmith:
Provide a master property to indicate if the current Routing Engine is the master. Provide a re_name property to return the name of the current Routing Engine.

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -218,12 +218,13 @@ class _Connection(object):
         The current Routing Engine is the RE to which the NETCONF session is
         connected.
 
-        Note: This property is based on new-style fact gathering and the
-              value of currently cached facts. If there is a chance the
-              mastership state may have changed since the facts were cached,
-              then dev.facts_refresh() should be invoked prior to checking
-              this property. If old-style fact gathering is in use,
-              this property will return None.
+        .. note::
+            This property is based on new-style fact gathering and the
+            value of currently cached facts. If there is a chance the
+            mastership state may have changed since the facts were cached,
+            then dev.facts_refresh() should be invoked prior to checking
+            this property. If old-style fact gathering is in use,
+            this property will return None.
 
         :returns: True if the current RE is the master Routing Engine. False if
                   the current RE is not the master Routing Engine. None if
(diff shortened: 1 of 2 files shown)


lib/jnpr/junos/device.py
```diff
@@ -206,6 +206,119 @@ def port(self):
...
+                else:
+                    for re_state in self.facts['current_re']:
+                        # Multi-chassis case. A chassis master/backup, but
+                        # not the system master/backup.
+                        if '-backup' in re_state or '-master' in re_state:
+                            master = False
+                            break
+        except KeyError:
+            # Old style fact gathering is likely in use. Nothing we can do.
+            pass
```
vnitinv: As per pep20 "Errors should never pass silently."

refer:https://www.python.org/dev/peps/pep-0020/

lib/jnpr/junos/device.py
```diff
@@ -206,6 +206,119 @@ def port(self):
...
+                elif 'backup' in self.facts['current_re']:
+                    master = False
+                # Some single RE platforms don't have 'master' in the
+                # 'current_re' fact. It's best to check if it's a single RE
+                # platform based on the '2RE' fact, not the 'current_re' fact.
+                elif (self.facts['2RE'] is False and
+                      're0' in self.facts['current_re']):
+                    master = True
+                else:
+                    for re_state in self.facts['current_re']:
```
vnitinv: Can you please explain me this for loop

lib/jnpr/junos/device.py
stacywsmith: I don't really consider this an "error". It's simply impossible to determine the master state if old-style fact gathering is in use. It's really similar to the situation where we return `None` for any facts which can't be gathered.

lib/jnpr/junos/device.py
```diff
@@ -206,6 +206,119 @@ def port(self):
...
+              this property will return None.
+
+        :returns: True if the current RE is the master Routing Engine. False if
+                  the current RE is not the master Routing Engine. None if
+                  unable to determine the state of the current Routing Engine.
+        """
+        master = None
+
+        try:
+            # Make sure the 'current_re' fact has a value
```
vnitinv: If this itself fails, then Master value will be None, Should we start by setting value to master as True in place of None?

lib/jnpr/junos/device.py
```diff
@@ -206,6 +206,119 @@ def port(self):
...
+                            re_name = re_state
+                            break
+                    # Still haven't figured it out, if there's only one key
+                    # in 'hostname_info', assume that.
+                    if (re_name is None and
+                       len(self.facts['hostname_info']) == 1):
+                        re_name = self.facts['hostname_info'].keys()[0]
+        except KeyError:
+            # Old style fact gathering is likely in use. Nothing we can do.
+            pass
```
vnitinv: same as above. It will be a very rare case that with the latest PyEZ user will use old-style facts gathering.

lib/jnpr/junos/device.py
stacywsmith: That was by design. As defined in the doc string, `None` is used to define that the mastership state could not be determined. I don't think we should default to assuming it is master if there's a problem determining the state.

lib/jnpr/junos/device.py
stacywsmith: I agree it will be a rare case, but one that I still think it's appropriate to handle.

lib/jnpr/junos/device.py
stacywsmith: Connecting to mj...
mj
In a multi-chassis system such as EX VC or TX Matrix, it is possible that an RE is neither master or backup. Here's the output of the current_re fact from a TX Matrix LCC:
```
   current_re: ['lcc1-re0', 'member1-re0', 'lcc1-master', 'member1-master', 'member1']
```

In this case, neither `'master'` or `'backup'` are present in `current_re`, and since these are multi-RE systems, the `2RE` fact is `True`. In this case, we check if any of the values in the `current_re` fact contain the string `-master` or `-backup`. That's an indication that the chassis is NOT the overall system master, and therefore we set `master = False` in this case.

PR #683, opened 2017-03-16
Version changes.

stacywsmith:
- Moving to version 2.1.dev0
- Adding support for internal version as a local version identifier
- Update jnpr.junos so it can be imported even if non-std libraries are not installed
- Update setup.py to pull version and author info from jnpr.junos.version.


stacywsmith:
@vnitinv Don't bother reviewing this yet. There's a problem related to namespace packages which is causing the installation failure. I need to figure that out before we even consider this.

PR #685, opened 2017-03-20
Enhance the ignore_warning decorator to check all warnings.

stacywsmith:
Instead of checking the first warning/error against the value of ignore_warning, check all warnings/errors against the value of ignore_warning.

### DIFF
--- a/lib/jnpr/junos/decorators.py
+++ b/lib/jnpr/junos/decorators.py
@@ -1,6 +1,7 @@
 # stdlib
 from functools import wraps
 import re
+import sys
 
 from jnpr.junos.exception import RpcError
 from jnpr.junos import jxml as JXML
@@ -117,7 +118,10 @@ def ignoreWarnDecorator(function):
                 if hasattr(ex, 'errs'):
                     for err in ex.errs:
                         if err['severity'] == 'warning':
-                            if isinstance(ignore_warn, (str, unicode)):
+                            if ((sys.version < '3' and
+                                 isinstance(ignore_warn, (str, unicode))) or
+                                (sys.version >= '3' and
+                                     isinstance(ignore_warn, str))):
                                 if not re.search(ignore_warn, err['message'],
                                                  re.I):
                                     # Message did not match.
--- a/lib/jnpr/junos/decorators.py
+++ b/lib/jnpr/junos/decorators.py
@@ -85,36 +85,54 @@ def ignoreWarnDecorator(function):
     For example::
         dev.rpc.get(ignore_warning=True)
         dev.rpc.get(ignore_warning='vrrp subsystem not running')
-        dev.rpc.get(ignore_warning=['vrrp subsystem not running', 'statement not found'])
+        dev.rpc.get(ignore_warning=['vrrp subsystem not running',
+                                    'statement not found'])
         cu.load(cnf, ignore_warning='statement not found')
 
-    :ignore_warning: It can take take boolean value or string or list of string.
-        if True, it will ignore all warning. If string, it will ignore warning if
-        the statement matches given string. If list of strings, it will try to check
-        if warning statement is from any of the given strings in the list:
+    :ignore_warning: It can take take boolean value or string or list of
+        string. If True, it will ignore all warning. If string, it will
+        ignore warning if the statement matches given string. If list of
+        strings, it will try to check if warning statement is from any of the
+        given strings in the list.
+
+    .. note::
+            When the value of ignore_warning is a string, or list of strings,
+            the string is actually used as a case-insensitive regular
+            expression pattern. If the string contains only alpha-numeric
+            characters, as shown in the above examples, this results in a
+            case-insensitive substring match. However, any regular expression
+            pattern supported by the re library may be used for more
+            complicated match conditions.
     """
     @wraps(function)
     def wrapper(*args, **kwargs):
-        if 'ignore_warning' in kwargs:
-            ignore_warn = kwargs.pop('ignore_warning')
+        ignore_warn = kwargs.pop('ignore_warning', False)
+        if ignore_warn:
             try:
                 result = function(*args, **kwargs)
                 return result
             except RpcError as ex:
                 ex.rpc_xml = JXML.remove_namespaces(ex.rpc_xml)
-                if hasattr(ex, 'rpc_error') and\
-                        ex.rpc_error['severity'] == 'warning':
-                    if ignore_warn is True:
-                        return ex.rpc_xml
-                    elif isinstance(ignore_warn, (str, unicode)):
-                        if re.search(ignore_warn, ex.message, re.I):
-                            return ex.rpc_xml
-                    elif isinstance(ignore_warn, list):
-                        for warn_msg in ignore_warn:
-                            if re.search(warn_msg, ex.message, re.I):
-                                return ex.rpc_xml
-                    raise ex
+                if hasattr(ex, 'errs'):
+                    for err in ex.errs:
+                        if err['severity'] == 'warning':
+                            if isinstance(ignore_warn, (str, unicode)):
+                                if not re.search(ignore_warn, err['message'],
+                                                 re.I):
+                                    raise ex
+                            elif isinstance(ignore_warn, list):
+                                for warn_msg in ignore_warn:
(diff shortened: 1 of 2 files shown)


lib/jnpr/junos/decorators.py
```diff
@@ -85,36 +85,54 @@ def ignoreWarnDecorator(function):
...
-                    elif isinstance(ignore_warn, list):
-                        for warn_msg in ignore_warn:
-                            if re.search(warn_msg, ex.message, re.I):
-                                return ex.rpc_xml
-                    raise ex
+                if hasattr(ex, 'errs'):
+                    for err in ex.errs:
+                        if err['severity'] == 'warning':
+                            if isinstance(ignore_warn, (str, unicode)):
+                                if not re.search(ignore_warn, err['message'],
```
vnitinv: I too was using re.search, but shouldn't we use re.match. Say in place of "vrrp subsystem not running" If I pass "xxxxvrrp subsystem not running" this too will pass, which should not be the case.

lib/jnpr/junos/decorators.py
```diff
@@ -85,36 +85,54 @@ def ignoreWarnDecorator(function):
...
+                                                 re.I):
+                                    raise ex
+                            elif isinstance(ignore_warn, list):
+                                for warn_msg in ignore_warn:
+                                    if not re.search(warn_msg, err['message'],
```
vnitinv: same as above.

vnitinv: I think this logic is wrong. say xml reply is

```
<rpc-error>
        <error-severity>warning</error-severity>
        <source-daemon>l2cpd</source-daemon>
        <error-path>[edit protocols]</error-path>
        <error-info>
            <bad-element>lldp</bad-element>
        </error-info>
        <error-message>Advertisement-interval is less than four</error-message>
    </rpc-error>
    <rpc-error>
        <error-severity>warning</error-severity>
        <source-daemon>chassisd</source-daemon>
        <error-path>[edit]</error-path>
        <error-info>
            <bad-element>chassis</bad-element>
        </error-info>
        <error-message>A system reboot is mandatory</error-message>
    </rpc-error>
```
errs is
```python
[{'severity': 'warning', 'edit_path': None, 'message': 'Advertisement-interval is less than four', 'bad_element':'l2cpd', 'source': ‘l2cpd’},
 {'severity': 'warning', 'edit_path': None, 'message': 'A system reboot is mandatory', 'bad_element': 'chassis', 'source': ‘chassisd’}]
```

if I give ignore_warning in different order
ignore_warning = ['A system reboot is mandatory’, 'Advertisement-interval is less than four']

'A system reboot is mandatory’ will be matched with 'Advertisement-interval is less than four' resulting in failure. Hence raising exception. 

**Note**: I have not tested this.

lib/jnpr/junos/decorators.py
stacywsmith: I think I prefer that we continue to use `re.search` instead of `re.match`.

1) `re.match` only checks at the beginning of each string, not the beginning of each line. Unfortunately, the message returned in the RPC reply sometimes has leading and trailing newlines (`\nstatement not found.\n` rather than just `statement not found.`)

2) `re.search` provides more capability and flexibility. For example, if you only want to match at the start of a line, you can do `ignore_warning='^vrrp subsystem not running'` However, I do think it's important that we properly document the behavior, and that's what I attempted to do with the doc string addition I made in this pull request. Here's the addition:
```
+    .. note::
 +            When the value of ignore_warning is a string, or list of strings,
 +            the string is actually used as a case-insensitive regular
 +            expression pattern. If the string contains only alpha-numeric
 +            characters, as shown in the above examples, this results in a
 +            case-insensitive substring match. However, any regular expression
 +            pattern supported by the re library may be used for more
 +            complicated match conditions.
```

lib/jnpr/junos/decorators.py
stacywsmith: Good catch. I'll fix this.

lib/jnpr/junos/decorators.py
```diff
@@ -119,14 +119,20 @@ def wrapper(*args, **kwargs):
                             if isinstance(ignore_warn, (str, unicode)):
                                 if not re.search(ignore_warn, err['message'],
```
vnitinv: If am bit concerned of using "if not"

say we have error return as
```
<rpc-error>
        <error-severity>warning</error-severity>
        <source-daemon>l2cpd</source-daemon>
        <error-path>[edit protocols]</error-path>
        <error-info>
            <bad-element>lldp</bad-element>
        </error-info>
        <error-message>Advertisement-interval is less than four</error-message>
    </rpc-error>
    <rpc-error>
        <error-severity>warning</error-severity>
        <source-daemon>chassisd</source-daemon>
        <error-path>[edit]</error-path>
        <error-info>
            <bad-element>chassis</bad-element>
        </error-info>
        <error-message>A system reboot is mandatory</error-message>
    </rpc-error>
```
if I give ignore_warning = "A system reboot is mandatory"

during 1st iteration, err['message'] will be "Advertisement-interval is less than four", hence search fails (so "if not" passes) and then throw the exception. 

Are we trying to say that if I am passing ignore_warning as 1 string, then the rpc-error reply from device should also have only one warning, otherwise if there are more than 1 warning are coming from device it should not ignore such response?

lib/jnpr/junos/decorators.py
```diff
@@ -119,14 +119,20 @@ def wrapper(*args, **kwargs):
...
                                     raise ex
                             elif isinstance(ignore_warn, list):
                                 for warn_msg in ignore_warn:
-                                    if not re.search(warn_msg, err['message'],
-                                                     re.I):
-                                        raise ex
+                                    if re.search(warn_msg, err['message'],
+                                                 re.I):
+                                        # Warning matches. Break skips else.
+                                        break
```
vnitinv: say we have errors as
```
[{'severity': 'warning', 'edit_path': None, 'message': 'Advertisement-interval is less than four', 'bad_element':'l2cpd', 'source': ‘l2cpd’},
 {'severity': 'warning', 'edit_path': None, 'message': 'A system reboot is mandatory', 'bad_element': 'chassis', 'source': ‘chassisd’}]
```
user pass
ignore_warning = ['A system reboot is mandatory', 'xxxx', 'yyyy']

here the device returned 2 warnings, one matches what being passed in the list, other "'Advertisement-interval is less than four'" is not being match. With a break statment, we are just ignoring 2nd warning. does our ignore_warning means that even if any of the warnings from Device matches any from the list give by user, We are good?

lib/jnpr/junos/decorators.py
stacywsmith: As I understood it, the expected behavior is that every warning message must match the value of `ignore_warning`. So, if the value of `ignore_warning` is a single string (such as `statement not found`, then every warning message in the response must match that value. If there are two warning messages with different values, and the user wants to ignore both of them, then they need to specify `ignore_warning` as a list of strings.

I believe my current code matches the behavior I've described above. If you see that it doesn't match this behavior or you disagree that this should be the expected behavior, then please do let me know.

lib/jnpr/junos/decorators.py
stacywsmith: In the case you just described, I would expect that the exception would be raised to the user because the first warning message (`Advertisement-interval is less than four`) did not match any of the `ignore_warning` values.

Stated another way, the normal exception is only suppressed if every warning message matches at least one of the values in `ignore_warning`.

I believe that's what my current code does. Here's a quick run down:
- If the exception doesn't have an `errs` attribute, raise the exception (line 142)
- Loop through every error/warning message (line 117):
    - If the message is not a warning, raise the exception (line 136)
    - If `ignore_warning` is a single string, and the message doesn't match, raise the exception (line 123)
    - If `ignore_warning` is a list of strings, loop through each string in the list:
        - Compare the message to the current string from `ignore_warning`
        - If the message matches the current string from `ignore_warning`, go to the next warning/error message
       - If the message didn't match any string in the `ignore_warning` list, then raise the exception (line 133)
- If all messages have been checked, then we know that all messages matched at least one of the values in `ignore_warning`, and therefore the exception should be suppressed and the rpc reply should be returned (line 138).

Again, let me know if this still doesn't make sense.

PR #691, opened 2017-03-27
Fix regressions caused by `ignore_warning`

stacywsmith:
In some circumstances, the implementation of the `ignore_warning` decorator didn't return the same value that would have been returned prior to the `ignore_warning` implementation. Once of those circumstances was when a `CommitError` was raised. The content of the `CommitError` was incorrectly empty.


vnitinv:

PR #700, opened 2017-04-17
SW() updates for new fact gathering. Fix Juniper/ansible-junos-stdlib#222

stacywsmith:
The SW class previously depended upon the existence of version_REx fact keys.
With enhanced fact gathering, these keys always exist, and are not a sign that
the device actually has multiple REs.

### DIFF
--- a/lib/jnpr/junos/utils/sw.py
+++ b/lib/jnpr/junos/utils/sw.py
@@ -62,7 +62,10 @@ class SW(Util):
         self._RE_list = []
         if 'junos_info' in dev.facts and dev.facts['junos_info'] is not None:
             self._RE_list = list(dev.facts['junos_info'].keys())
-        self._multi_RE = bool(dev.facts['2RE'])
+        else:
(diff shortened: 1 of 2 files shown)


vnitinv:

vnitinv:
ok to test

lib/jnpr/junos/utils/sw.py
```diff
@@ -59,9 +59,10 @@ class SW(Util):
     def __init__(self, dev):
         Util.__init__(self, dev)
         self._dev = dev
-        self._RE_list = [
-            x for x in dev.facts.keys() if x.startswith('version_RE')]
-        self._multi_RE = bool(len(self._RE_list) > 1)
+        self._RE_list = []
+        if 'junos_info' in dev.facts and dev.facts['junos_info'] is not None:
```
vnitinv: Is this code backward compatible? I think we dont have 'junos_info' earlier hence in absence of else self._RE_list will be a empty list only.
Please check

Issue #701, opened 2017-04-18
Invalid JSON generated with rpc.get_config(options={'format': 'json'})


CMeza99:
```python
>>>json.loads(dev.rpc.get_config(options={'format': 'json'}).text)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/json/__init__.py", line 319, in loads
    return _default_decoder.decode(s)
  File "/usr/local/lib/python3.5/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/lib/python3.5/json/decoder.py", line 355, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 22030 column 17 (char 680605)
```

Looking at the json from `dev.rpc.get_config(options={'format': 'json'}).text` it indeed is missing a `,`:
```vim
22021         "protocols" : [
22022         {
22023             "comment" : "BEGIN: ROUTING-INSTANCES",
22024             "bgp" : [
22025             {
22026                 "apply-groups" : [
22027                 {
22028                     "data" : "BGP_DEFAULTS"
22029                 }
22030                 ]
22031                 "path-selection" : [
22032                 {
22033                     "always-compare-med" : [
22034                     {
22035                         "data" : null
22036                     }
22037                     ]
22038                 }
22039                 ],
```

stacywsmith:
Hi @CMeza99 

Thanks for reporting. This appears to be a Junos bug in the rendering of the configuration in JSON format. If you want to provide more details here I can help in tracking the Junos issue. Alternatively, you can open a JTAC case.

I would need to know:
1) Juniper platform/model
2) Exact version of Junos
3) CLI output of `show configuration | display xml`
4) CLI output of `show configuration | display json`

FYI, you should not have to call `json.loads()` on the text of the RPC reply. Instead, when you request JSON format, PyEZ should perform the `json.loads()` for you and return a native Python data structure.

In this case, that's not happening because the `json.loads()` is raising an exception because the JSON emitted by Junos is invalid. I spoke with @vnitinv about this and he will address this specific PyEZ issue.

CMeza99:
Thank you
1. QFX10008 and QFX10000-RE
2. Junos 15.1X53-D36
[...]

PR #706, opened 2017-04-20
raise JSONLoadError if json config is malformed

vnitinv:
Handle error case for https://github.com/Juniper/py-junos-eznc/issues/701

exception would look like

```python
jnpr.junos.exception.JSONLoadError: JSONLoadError(reason: Expecting , delimiter: line 461 column 39 (char 16862), 
The offending config appears to be: 
459:                         "file" : {
460:                             "filename" : "netconf-ops.log", 
461:                             "size" : 3m, 
462:                             "files" : 20, 
463:                             "world-readable" : [null]
```

### DIFF
--- a/lib/jnpr/junos/exception.py
+++ b/lib/jnpr/junos/exception.py
@@ -1,3 +1,4 @@
+import re
 from jnpr.junos import jxml
 from jnpr.junos import jxml as JXML
 from lxml.etree import _Element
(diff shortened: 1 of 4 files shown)


lib/jnpr/junos/exception.py
```diff
@@ -297,3 +298,26 @@ class ConnectClosedError(ConnectError):
...
+        self.offending_line = ''
+        obj = re.search('line (\d+)', self.ex_msg)
+        if obj:
+            line_no = int(obj.group(1))
+            self.offending_line = rpc_content.splitlines()[line_no-1]
+
+    def __repr__(self):
+        return "{0}(reason: {1}, \nThe offending config appears to be: \n{2}" \
+                .format(self.__class__.__name__, self.ex_msg,
+                        self.offending_line)
```
stacywsmith: I like your idea of providing 5 lines (2 pre + 1 + 2 post) instead of just the single offending line. In addition, I suggest you print these 5 lines with line numbers prepended to the actual config line.

jnpr-community-netdev: Done the changes as discussed.

vnitinv: Done the changes as discussed.

PR #707, opened 2017-04-21
ephemeral config support

vnitinv:
Now we can configure using ephemeral db

sample code:
```python
with Config(dev, mode='ephemeral', ephemeral_instance='xyz') as cu:
    cnf = """<configuration>
                <protocols>
                    <ospf>
                        ---------
                    </ospf>
                </protocols>
        </configuration>"""
    cu.load(cnf)
    cu.commit()
```
for default ephemeral instance, we can just do:
```python
with Config(dev, mode='ephemeral') as cu:
    cnf = """<configuration>
                <protocols>
                    <ospf>
                        ---------
                    </ospf>
                </protocols>
        </configuration>"""
    cu.load(cnf)
    cu.commit()
```


lib/jnpr/junos/utils/config.py
```diff
@@ -766,15 +770,27 @@ def _open_configuration_batch():
...
+        else:
+            ephemeral_kwargs = {'ephemeral': True}
+
         {
             'private': _open_configuration_private,
             'dynamic': _open_configuration_dynamic,
             'batch': _open_configuration_batch,
-            'exclusive': _open_configuration_exclusive
+            'exclusive': _open_configuration_exclusive,
+            'ephemeral': lambda: _open_configuration_ephemeral(
```
stacywsmith: Can you explain why lambda is required here, but not in the other cases? Is it because `_open_configuration_ephemeral()` takes arguments?

vnitinv: yes

vnitinv:
I opted for kwargs in __init__() to take ephemeral_instance rather than using it directly because ephemeral is quite risky when not used properly. I wanted to hide the parameter for those reasons.
I can change it to be explicit.

stacywsmith:
@vnitinv Rather than hiding it with kwargs (which has side effects), I'd rather see a very explicit warning that says something like:
```
 Ephemeral databases are an advanced Junos feature which if used incorrectly can have serious negative impact on the operation of the Junos device. We recommend you consult JTAC and/or you Juniper account team before deploying the ephemeral database feature in your network.
```

vnitinv:
@stacywsmith Added warning message and kwargs check

vnitinv:
function test case failure is due to https://github.com/Juniper/py-junos-eznc/issues/717
will be fixed under different pull request.

PR #708, opened 2017-05-02
Handle ConnectClosedError exception for lock() and unlock()

stacywsmith:
It is possible that a ConnectClosedError exception is raised in
response to a lock() or unlock() call. This was being mishandled
as a direct error from ncclient.


vnitinv:

PR #711, opened 2017-05-04
Add a srx_cluster_redundancy_group fact.

stacywsmith:
The XML output of the SRX cluster redundancy group information is
not easy to parse. Presenting this information in a PyEZ fact makes
it accessible to most PyEZ users.

### DIFF
--- a/lib/jnpr/junos/facts/get_chassis_cluster_status.py
+++ b/lib/jnpr/junos/facts/get_chassis_cluster_status.py
@@ -8,7 +8,21 @@ def provides_facts():
     """
     return {'srx_cluster': "A boolean indicating if the device is part of an "
                            "SRX cluster.",
-            'srx_cluster_id': "A string containing the configured cluster id"}
+            'srx_cluster_id': "A string containing the configured cluster id",
+            'srx_cluster_redundancy_group': "A multi-level dictionary of "
+                                            "information about the SRX "
+                                            "cluster redundancy groups "
+                                            "on the device. The first-level "
+                                            "key is the redundancy group id. "
+                                            "The second-level keys are: "
+                                            "cluster_id, failover_count, "
+                                            "node0, and node1. The node0 and "
+                                            "node1 keys have third-level keys "
+                                            "of priority, preempt, status, "
+                                            "and failover_mode. The values "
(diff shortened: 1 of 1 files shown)


vnitinv:

lib/jnpr/junos/facts/get_chassis_cluster_status.py
```diff
@@ -17,20 +31,45 @@ def get_facts(device):
     """
     srx_cluster = None
     srx_cluster_id = None
+    redundancy_group = None
 
     try:
-        rsp = device.rpc.get_chassis_cluster_status(normalize=True,
-                                                    redundancy_group="0")
+        rsp = device.rpc.get_chassis_cluster_status(normalize=True)
```
vnitinv: If by this time if we already have the facts with personality details, shouldnt we use that to check if the device type is SRX (which is not a major percentage of our products in customer place), hence saving unnecessary RPC calls to different type of devices.

stacywsmith: @vnitinv I see a few issues with your suggestion.

1) Because fact gathering is now done on demand, I feel there is little benefit from making the RPC call conditional on the personality. As long as the fact SRX-cluster-related facts aren't accessed, this RPC isn't called.

2) The new fact gathering system differentiates between `False` and `None` return values. A return value of `False` means the fact was successfully gathered and it's value is not true. A value of `None` means the fact could not be gathered definitively. If we don't call this RPC, then we can't definitively gather this particular fact. We would potentially be returning a different values than is returned with the current. (See final point.)

3) Right now, the `personality` fact is some of the most fragile fact gathering code. It makes some assumptions that may not continue to hold true as new product models are introduced. Right now a 3-digit SRX model is assumed to be `'SRX_BRANCH'` and a 4-digit SRX model is assumed to be `'SRX_HIGHEND'`. For example, if a 5-digit SRX model is introduced, then the `personality` will be `None`. While that's obviously something that could be updated in the gathering of the `personality` fact, I'd rather not make the other facts dependent upon this fragile fact.

Based on these points, I recommend we not make the RPC call dependent upon the `personality` fact.

PR #712, opened 2017-05-04
ignore_warning fails when single <rpc-error> that is first child of <rpc-reply>.

stacywsmith:
The implementation of the ignore_warning decorator failed when there was
a single warning `<rpc-error>` element that was the first child of the
`<rpc-reply>` element, such as this example:
```
<rpc-reply xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" xmlns:junos="http://xml.juniper.net/junos/16.1R4/junos" xmlns:nc="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="urn:uuid:65b27597-c62d-4e8b-ba85-1ca000646a9e">
    <load-configuration-results>
        <rpc-error>
            <error-severity>warning</error-severity>
            <error-path>[edit]</error-path>
            <error-message>statement not found: system</error-message>
        </rpc-error>
        <ok/>
    </load-configuration-results>
</rpc-reply>
```

### DIFF
--- a/lib/jnpr/junos/decorators.py
+++ b/lib/jnpr/junos/decorators.py
@@ -117,53 +117,52 @@ def ignoreWarnDecorator(function):
         except RPCError as ex:
             if hasattr(ex, 'xml') and ignore_warning:
                 if hasattr(ex, 'errors'):
-                    for err in ex.errors:
-                        if err.severity == 'warning':
-                            if ((sys.version < '3' and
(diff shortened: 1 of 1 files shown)


vnitinv:

stacywsmith:
@vnitinv Can you please review? The changes are not as extensive as the diff indicates. It's really just lines 120-122 and 155 which are changed.

PR #713, opened 2017-05-04
mode='telnet' did not logout non-cli user

stacywsmith:
When connecting to the device via console server with mode='telnet',
if the login user was configured to log into the shell, like the root
user, then the session was not logged out completely upon close.

### DIFF
--- a/lib/jnpr/junos/transport/tty.py
+++ b/lib/jnpr/junos/transport/tty.py
@@ -50,7 +50,7 @@ class Terminal(object):
         _re_pat_login,
         '(?P<passwd>assword:\s*$)',
         '(?P<badpasswd>ogin incorrect)',
-        '(?P<already_closed>session end at .*\n%)',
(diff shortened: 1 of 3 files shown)


vnitinv:

stacywsmith:
This pull request fixes #714.

lib/jnpr/junos/transport/tty.py
```diff
@@ -50,7 +50,7 @@ class Terminal(object):
         _re_pat_login,
         '(?P<passwd>assword:\s*$)',
         '(?P<badpasswd>ogin incorrect)',
-        '(?P<already_closed>session end at .*\n%)',
+        '(?P<netconf_closed>\<\!\-\- session end at .*\-\-\>\s*)',
```
vnitinv: We only need to escape control characters, (+ ? . * ^ $ ( ) [ ] { } | \), So why did we opt to use escape for "!,<,>,-"

stacywsmith: Very valid point. I will fix. Thanks!

PR #718, opened 2017-05-12
JSONLoadError was thrown when load valid JSON config

### LINKED ISSUE #717: JSONLoadError even when JSON config being loaded in proper
```python
    cnf = """{
        "configuration" : {
            "system" : {
                "services" : {
                    "telnet" : [null]
                 }
            }
        }
    }"""
    with Config(dev) as cu:
        cu.load(cnf, format='json')
```
throw JSONLoadError even when JSON config being loaded in proper
This issue got introduced after the changes done for enhancement under https://github.com/Juniper/py-junos-eznc/pull/706

vnitinv:
fix #717


vnitinv:
ok to test

PR #721, opened 2017-05-16
Return None when the RPC equiv is either str or unicode

mirceaulinic:
Currently it checks against str, and it borks when the command
is sent as unicode. An example of issue is under
https://github.com/napalm-automation/napalm-junos/issues/152.
To make sure it works with both, we need to check against
the base string class.


jnpr-community-netdev:

stacywsmith:
Thanks for the fix @mirceaulinic

PR #722, opened 2017-05-23
Handle differences in |display xml rpc

stacywsmith:
On some devices, dev.rpc.cli() the use of `| display xml rpc` requires the explicit setting of `format="xml"`. In addition, on some devices, the response may not be wrapped in `<rpc-reply>` tags. This change handles both of those issues while remaining backwards compatible.


jnpr-community-netdev:

PR #728, opened 2017-06-02
Optimize image copying in SW.safe_copy()

stacywsmith:
The SW.install() method uses the SW.safe_copy() method to copy
a software package to a Junos device if the no_copy argument is False.

This enhancement checks to see if the package is already present on the
Junos device and skips the copy if the package is present and the
file checksum of botht the local and remote packages match.

Also added additional arguments to control the copy and checksum
functionality.

### DIFF
--- a/lib/jnpr/junos/utils/sw.py
+++ b/lib/jnpr/junos/utils/sw.py
@@ -150,7 +150,7 @@ class SW(Util):
 
     def put(self, package, remote_path='/var/tmp', progress=None):
         """
-        SCP or FTP 'put' the package file from the local server to the remote 
+        SCP or FTP 'put' the package file from the local server to the remote
         device.
 
         :param str package:
@@ -453,8 +453,7 @@ class SW(Util):
                                  (algorithm))
             return rsp.findtext('.//checksum')
         except RpcError as e:
-            if (hasattr(e, 'message') and
-               'No such file or directory' in e.message):
+            if 'No such file or directory' in getattr(e, 'message', ''):
                 return None
             else:
                 raise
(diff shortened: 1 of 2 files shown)


jnpr-community-netdev:

lib/jnpr/junos/utils/sw.py
```diff
@@ -392,36 +417,44 @@ def _issu_nssu_requirement_validation(self):
...
+                raise ValueError('Unknown checksum algorithm: %s' %
+                                 (algorithm))
+            return rsp.findtext('.//checksum')
         except RpcError as e:
-
-            # e.errs is list of dictionaries
-            if hasattr(e, 'errs') and \
-                    list(filter(lambda x: 'No such file or directory' in
-                                x['message'], e.errs)):
+            if (hasattr(e, 'message') and
```
vnitinv: in python 3.x, hasattr(e, 'message') might always give False

lib/jnpr/junos/utils/sw.py
```diff
@@ -444,53 +480,89 @@ def safe_copy(self, package, **kvargs):
...
         if checksum is None:
-            _progress('computing local checksum on: %s' % package)
-            checksum = SW.local_md5(package)
+            _progress('computing checksum on local package: %s' % (package))
+            try:
+                checksum = SW.local_checksum(package,
+                                             algorithm=checksum_algorithm)
+            except IOError:
+                _progress('error computing checksum on local package: %s. '
+                          'Ensure the local package exists.' % (package))
```
vnitinv: should we not return False from this exception?

stacywsmith: Yes. Thanks. Fixed.

Issue #723, opened 2017-05-25
Cannot pack pyez-script into .exe file by Pyinstaller


tongorz:
The error output from pyinstaller is like:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/_MEIRsuH58/jnpr/junos/facts'

Same issue page under pyinstaller:
[https://github.com/pyinstaller/pyinstaller/issues/2438](url)

Answer given by pyinstaller is that:

Please ask the vendor of that package to fix this issue in their code. Proper solutions would be:
1. name all sub-modules in jnpr/junos/facts/__init__.py (this is the easiest one)
2. when running setup.py write existing sub-module's names into a text file and read it using pkg_resources (flexible, but most complicated to implement), or
3. use the pkg_resources' plugin-in mechanism.

vnitinv:
Thanks for your suggestion @tongorz . Fixed via #731

PR #731, opened 2017-06-09
Explicitly initialize jnpr.junos.facts sub-modules. Addresses #723

stacywsmith:
Previously, the sub-modules within jnpr/junos/facts/ were dynamically
loaded based on their filename. However, that conflicted with some
tools such as Pyinstaller.

Changed to explicitly load the sub-modules.


jnpr-community-netdev:

Issue #735, opened 2017-06-20
Cryptic error if timeout is a string value


stacywsmith:
The `dev.timeout` attribute is passed through to ncclient.

If the `dev.timeout` attribute is set to a non-integer value, the following cryptic error message is produced from ncclient: 

```
"/usr/local/python-3.4.5/lib/python3.4/site-packages/jnpr/junos/rpcmeta.py", line 336, in _exec_rpc return self._junos.execute(rpc, **dec_args) File "/usr/local/python-3.4.5/lib/python3.4/site-packages/jnpr/junos/decorators.py", line 76, in wrapper return function(*args, **kwargs) File "/usr/local/python-3.4.5/lib/python3.4/site-packages/jnpr/junos/decorators.py", line 23, in wrapper result = function(*args, **kwargs) File "/usr/local/python-3.4.5/lib/python3.4/site-packages/jnpr/junos/device.py", line 712, in execute ignore_warning=ignore_warning) File "/usr/local/python-3.4.5/lib/python3.4/site-packages/jnpr/junos/decorators.py", line 116, in wrapper rsp = function(self, *args, **kwargs) File "/usr/local/python-3.4.5/lib/python3.4/site-packages/jnpr/junos/device.py", line 1260, in _rpc_reply return self._conn.rpc(rpc_cmd_e)._NCElement__doc File "/usr/local/python-3.4.5/lib/python3.4/site-packages/ncclient/manager.py", line 172, in wrapper return self.execute(op_cls, *args, **kwds) File "/usr/local/python-3.4.5/lib/python3.4/site-packages/ncclient/manager.py", line 232, in execute raise_mode=self._raise_mode).request(*args, **kwds) File "/usr/local/python-3.4.5/lib/python3.4/site-packages/ncclient/operations/third_party/juniper/rpc.py", line 44, in request return self._request(rpc) File "/usr/local/python-3.4.5/lib/python3.4/site-packages/ncclient/operations/rpc.py", line 322, in _request self._event.wait(self._timeout) File "/usr/local/python-3.4.5/lib/python3.4/threading.py", line 546, in wait signaled = self._cond.wait(timeout) File "/usr/local/python-3.4.5/lib/python3.4/threading.py", line 293, in wait if timeout > 0:
```

This can occur if `dev.timeout` is directly set to a non-integer value, or if the value of the `dev_timeout` decorator on an RPC call is set to a non-integer value.

vnitinv:
Fixed via #736

PR #736, opened 2017-06-20
Ensure dev.timeout is an integer value. Addresses #735

stacywsmith:
Try to convert the dev.timeout value to an integer to ensure
that ncclient receives an integer value. If the value can not
be converted to an integer, raise a RuntimeError with a meaningful
error message.


jnpr-community-netdev:

vnitinv:
ok to test

vnitinv:
fix #735

vnitinv:
ok to test

PR #740, opened 2017-06-23
Support platforms which have single-RE ISSU.

stacywsmith:
Support platforms which have single-RE ISSU.
Handle the fact that some platforms don't return <package-result> from sw install.
Unit test fixup and PEP8 cleanup.

### DIFF
--- a/lib/jnpr/junos/utils/sw.py
+++ b/lib/jnpr/junos/utils/sw.py
@@ -243,7 +243,9 @@ class SW(Util):
 
     def _parse_pkgadd_response(self, rsp):
         got = rsp.getparent()
-        rc = int(got.findtext('package-result').strip())
+        # If <package-result> is not present, then assume success.
+        # That is, assume <package-result>0</package-result>
+        rc = int(got.findtext('package-result','0').strip())
         output_msg = '\n'.join([i.text for i in got.findall('output')
                                 if i.text is not None])
         self.log("software pkgadd package-result: %s\nOutput: %s" % (
--- a/lib/jnpr/junos/utils/sw.py
+++ b/lib/jnpr/junos/utils/sw.py
@@ -245,7 +245,7 @@ class SW(Util):
         got = rsp.getparent()
         # If <package-result> is not present, then assume success.
         # That is, assume <package-result>0</package-result>
-        rc = int(got.findtext('package-result','0').strip())
+        rc = int(got.findtext('package-result', '0').strip())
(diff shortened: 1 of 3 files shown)


stacywsmith:
ok to test

jnpr-community-netdev:

lib/jnpr/junos/utils/sw.py
```diff
@@ -241,7 +243,9 @@ def pkgaddISSU(self, remote_package, **kvargs):
 
     def _parse_pkgadd_response(self, rsp):
         got = rsp.getparent()
-        rc = int(got.findtext('package-result').strip())
+        # If <package-result> is not present, then assume success.
+        # That is, assume <package-result>0</package-result>
```
vnitinv: Is this valid assumption?

lib/jnpr/junos/utils/sw.py
```diff
@@ -70,6 +70,8 @@ def __init__(self, dev):
             self._multi_RE is True and dev.facts.get('vc_capable') is True and
             dev.facts.get('vc_mode') != 'Disabled')
         self._mixed_VC = bool(dev.facts.get('vc_mode') == 'Mixed')
+        self._single_re_issu = bool('current_re' in dev.facts and
+                                    'localre' in dev.facts['current_re'])
```
vnitinv: Can I get any doc reference for "localre"?

lib/jnpr/junos/utils/sw.py
stacywsmith: I can't guarantee it's a perfect assumption, but I do believe it's valid. In my experience, platforms which do not include `<package-result>` will include an `<rpc-error>` element if there's an error. We'll raise an `RpcError` exception in that case.

lib/jnpr/junos/utils/sw.py
stacywsmith:
```
{master:0}
user@s0> file show /etc/hosts.junos | match localre 
128.0.0.63       localre
```
The platforms which support single-RE ISSU use this internal IP address for communication during the ISSU process.

stacywsmith:
I'm fine waiting until after 2.1.4 release to merge this.

lib/jnpr/junos/utils/sw.py
vnitinv: @stacywsmith Please add a log for users reference when <package-result> is missing

stacywsmith: @vnitinv Log added.

PR #746, opened 2017-06-27
Single-RE sw install on multi-RE device using all_re argument to SW.install()

stacywsmith:


jnpr-community-netdev:

lib/jnpr/junos/utils/sw.py
```diff
@@ -783,8 +789,8 @@ def _progress(report):
                     "NSSU: installing software ... please be patient ...")
                 return self.pkgaddNSSU(remote_package,
                                        dev_timeout=timeout, **kwargs)
-            elif self._multi_RE is False:
-                # simple case of device with only one RE
+            elif self._multi_RE is False or all_re is False:
```
vnitinv: @stacywsmith I understand with this change we dont have any code coverage issue, but can we still add a UT.

stacywsmith: @vnitinv Yes. Valid point. I will add a UT.

PR #749, opened 2017-06-29
load config using <load-configuration> RPCs url attribute

vnitinv:
Now user can load config using `url` attribute. with this user can load config file located on the device itself.
It should also support all the options with url as per [doc](https://www.juniper.net/documentation/en_US/junos/topics/reference/tag-summary/junos-xml-protocol-load-configuration.html)

Example:
```python
  cu.load(url="/var/home/user/golden.conf")
  cu.load(url="ftp://username@ftp.hostname.net/filename")
  cu.load(url="http://username:password@hostname/path/filename")
  cu.load(url="/var/home/user/golden.conf", overwrite=True)
```

### DIFF
--- a/lib/jnpr/junos/utils/config.py
+++ b/lib/jnpr/junos/utils/config.py
@@ -18,7 +18,7 @@ Configuration Utilities
 
 class Config(Util):
     """
-    Overivew of Configuration Utilities.
+    Overview of Configuration Utilities.
 
     * :meth:`commit`: commit changes
     * :meth:`commit_check`: perform the commit check operation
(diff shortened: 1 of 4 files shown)


lib/jnpr/junos/utils/config.py
```diff
@@ -17,9 +17,8 @@
 
 
 class Config(Util):
-
     """
-    Overivew of Configuration Utilities:
+    Overivew of Configuration Utilities.
```
vnitinv: Done

vnitinv:
@stacywsmith I have done the changes as per review comment. Please do check it once again. Thanks

PR #751, opened 2017-06-29
SW.install() enhanced to install from a URL.

stacywsmith:
The `SW.install()` method required either a `package` or `pkg_set` argument
to specify the source path (on the PyEZ host's local filesystem) of the
software package. The package was then copied from the PyEZ host to the
`remote_path` directory of the target device. The the software was
installed from the file which had been copied to the target device.

Junos also supports the ability to specify the software to be installed
using a URL rather than a file on the device's local filesystem.

This enhancement adds that capability to the `SW.install()` method by
specifying the `url` argument rather than the `package` or `pkg_set`
argument.

i.e.
```
result = sw.install(url='http://server.example.com/junos-install-mx-x86-64-16.1R4.7.tgz')
```

URLs can be specified using the formats documented [here](https://www.juniper.net/documentation/en_US/junos/topics/concept/junos-software-formats-filenames-urls.html).


jnpr-community-netdev:

stacywsmith:
@vnitinv I reworked the pull request to address your review comment. Since the code is completely changed from the previous version of the pull request, I'd appreciate it if you could review it again.

lib/jnpr/junos/utils/sw.py
```diff
@@ -595,27 +601,28 @@ def install(self, package=None, pkg_set=None, remote_path='/var/tmp',
         Performs the complete installation of the **package** that includes the
         following steps:
 
-        1. computes the checksum of :package: on the local host
+        1. If :url: is specified, or :no_copy: is True, skip to step 8.
```
vnitinv: this need to be edited, For example "if package specified is a url"

lib/jnpr/junos/utils/sw.py
```diff
@@ -757,41 +777,48 @@ def _progress(report):
...
         # at this point, the file exists on the remote device
+        # or will be loaded directly from a URL.
         # ---------------------------------------------------------------------
-        if package is not None:
-            remote_package = remote_path + '/' + path.basename(package)
-            if validate is True:  # in case of Mixed VC it cant be used
+
+        if len(remote_pkg_set) == 1:
+            # validate can't be used in the case of a Mixed VC
+            if validate is True and self._mixed_VC is False:
```
vnitinv: How about initializing 
remote_package = remote_pkg_set[0]

stacywsmith:
@vnitinv Incorporated the two minor changes you requested.

Issue #750, opened 2017-06-29
facts RE uptime with virtual chassis where RE fpc number > 1


lampwins:
The facts output includes `RE0` and `RE1` which are really intended for cases when the device has dual routing engines in a single chassis. This does not play well with a virtual chassis stack when the routing engines are in fpc positions greater than either 0 or 1. 

This came about because I wanted to obtain the uptime for a routing engine in the stack. The `uptime` is included the RE dicts when they are available in `RE0` and `RE1`. `re_info['default']` contains all the same information _except_ for uptime. I would like to have `uptime` added to the RE dicts in `re_info` to solve this issue.

stacywsmith:
`uptime` was intentionally excluded from the new `re_info` fact because it's not really a "fact". I consider "facts" to be read-only information which maintains the same value for the life of the PyEZ connection to the device. However, the value of uptime obviously changes over time. Even worse, the uptime value would always be the time at which the `re_info`, `RE0`, or `RE1` fact was retrieved, rather than the current uptime.

Of course, this is true in the older `RE0` and `RE1` facts as well, but the decision was made to maintain the current behavior of those facts for backward compatibility.

It's also worth noting that on-demand fact gathering makes things slightly less predictable. Previously, all facts were gathered when `open()` was invoked, or when `facts_refresh()` was invoked. Now, each fact is generally gathered the first time it's value is accessed. However, multiple facts might be derived from the same RPC, or one fact might be dependent upon another fact. In those cases, the fact is gathered the first time a dependent fact, or a fact derived from the same RPC, is gathered.

I would certainly be supportive of adding an `uptime()` Device method which always executes a fresh RPC to return the current uptime for all Routing Engines.

Would an `uptime()` method meet your requirement?

lampwins:
@stacywsmith thank you for the background. With that explanation, yes I fully support an `uptime()` which returns the current uptime for each RE.

stacywsmith:
@lampwins Would it be sufficient to always return the uptime of the current Routing Engine (the one the NETCONF connection is to) as a number of seconds, rather than returning a dictionary with the uptime of each RE?

lampwins:
@stacywsmith yes that is sufficient for my needs.

stacywsmith:
@lampwins Please take a look at #752 and let me know if you have any feedback.

stacywsmith:
Addressed via #752

lampwins:
[...]

PR #752, opened 2017-06-30
Implement uptime property for Device instances. Addresses #750.

stacywsmith:
Add a new `uptime` property which provides the (int) number of seconds since the current Routing Engine was booted.

```
>>> from jnpr.junos.device import Device
>>> dev = Device(host='r0', user='user', passwd='user123')
>>> dev.open()
Device(r0)
>>> dev.uptime
69436
>>> 
```

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -290,15 +290,14 @@ class _Connection(object):
         The current Routing Engine is the RE to which the NETCONF session is
         connected.
 
-        :returns: The number of seconds (int) the current Routing Engine has 
-                  been up. If there is a problem gathering information, None is
-                  returned.
+        :returns: The number of seconds (int) since the current Routing Engine
+                  was booted. If there is a problem gathering or parsing the
+                  uptime information, None is returned.
         :raises: May raise a specific jnpr.junos.RpcError or 
                  jnpr.junos.ConnectError subclass if there is a problem
                  communicating with the device.
         """
         uptime = None
-
         rsp = self.rpc.get_system_uptime_information(normalize=True)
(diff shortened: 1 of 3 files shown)


jnpr-community-netdev:

lib/jnpr/junos/device.py
```diff
@@ -279,6 +279,40 @@ def master(self, value):
...
+        connected.
+
+        :returns: The number of seconds (int) since the current Routing Engine
+                  was booted. If there is a problem gathering or parsing the
+                  uptime information, None is returned.
+        :raises: May raise a specific jnpr.junos.RpcError or
+                 jnpr.junos.ConnectError subclass if there is a problem
+                 communicating with the device.
+        """
+        uptime = None
```
vnitinv: As the return type is int, should we initialize this with 0?

stacywsmith: If for some reason the expected XML element doesn't exist, I think it makes more sense to return `None` than `0`.  This is also documented in the property:
```
If there is a problem gathering or parsing the  uptime information, None is returned.
```

The `None` value is also consistent with the `master` property and the facts in `dev.facts`. All of those return `None` if the information is unable to be gathered from the device.

vnitinv:
What do we lose by making this property attribute as part of facts itself. facts are static data but now as we have facts on demand, can't we change this rule?

stacywsmith:
@vnitinv I have a strong preference not to make this part of facts. You are, of course, technically correct that the current implementation of fact gathering could be modified not to cache certain facts which always required refreshing. However, I feel that would cause user confusion if we were to cache some "fact" values and didn't cache others.

PR #758, opened 2017-07-11
Properly handle normalize argument to the open() Device method. Fixes #757

### LINKED ISSUE #757: normalize argument to Device open() method not handled correctly
If the `normalize` argument is specified to the `open()` method, then
later RPC invocations which attempted to override the `normalize` value
might not work correctly.

This happens because `normalizeDecorator()` depends on the value of `dev._normalize`,
but `open()` is not properly setting the value of `dev._normalize`.

Here is an example where the default behavior of `normalize=False` is shown:
```
from jnpr.junos.device import Device
from lxml import etree

dev = Device(host='r0', user='user', passwd='user123')
dev.open()
etree.dump(dev.rpc.get_interface_information(interface_name='fxp0.0', terse=True))
dev.close()
```

The output from executing this script is:
```
<interface-information style="terse">
<logical-interface>
<name>
fxp0.0
</name>
<admin-status>
up
</admin-status>
<oper-status>
up
</oper-status>
<filter-information>
</filter-information>
<address-family>
<address-family-name>
inet
</address-family-name>
<interface-address>
<ifa-local emit="emit">
10.48.22.14/22
</ifa-local>
</interface-address>
</address-family>
</logical-interface>
</interface-information>
```
In this case, the output is not normalized. This is the correct and expected behavior for this case.

If we now specify `dev.open(normalize=True)`:
```
from jnpr.junos.device import Device
from lxml import etree

dev = Device(host='r0', user='user', passwd='user123')
dev.open(normalize=True)
etree.dump(dev.rpc.get_interface_information(interface_name='fxp0.0', terse=True))
dev.close()
```

The output from executing this script is:
```
<interface-information style="terse">
  <logical-interface>
    <name>fxp0.0</name>
    <admin-status>up</admin-status>
    <oper-status>up</oper-status>
    <filter-information/>
    <address-family>
      <address-family-name>inet</address-family-name>
      <interface-address>
        <ifa-local emit="emit">10.48.22.14/22</ifa-local>
      </interface-address>
    </address-family>
  </logical-interface>
</interface-information>
```
In this case, the output is normalized. Again, this is the correct and expected behavior for this case.

If we now add `normalize=False` to the RPC invocation:
```
from jnpr.junos.device import Device
from lxml import etree

dev = Device(host='r0', user='user', passwd='user123')
dev.open(normalize=True)
etree.dump(dev.rpc.get_interface_information(interface_name='fxp0.0', terse=True, normalize=False))
dev.close()
```

The output from executing this script is:
```
<interface-information style="terse">
  <logical-interface>
    <name>fxp0.0</name>
    <admin-status>up</admin-status>
    <oper-status>up</oper-status>
    <filter-information/>
    <address-family>
      <address-family-name>inet</address-family-name>
      <interface-address>
        <ifa-local emit="emit">10.48.22.14/22</ifa-local>
      </interface-address>
    </address-family>
  </logical-interface>
</interface-information>
```
In this case, the output is normalized. This is incorrect behavior. The `normalize=False` argument to the RPC invocation should have overridden the `normalize=True` argument to `open()`.

stacywsmith:
If the `normalize` argument was specified to the `open()` method, then
later RPC invocations which attempted to override the `normalize` value
might not work correctly.

This was because `normalizeDecorator()` depends on the value of `dev._normalize`,
but `open()` was not properly setting the value of `dev._normalize`.

This change properly sets the value of `dev._normalize` when the
`normalize` argument is specified in the call to the `open()` device method.

Also removed unnecessary commented out code in console.py.

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -1278,9 +1278,7 @@ class Device(_Connection):
         # normalize argument to open() overrides normalize argument value
         # to __init__(). Save value to self._normalize where it is used by
         # normalizeDecorator()
-        normalize = kvargs.get('normalize')
-        if normalize is not None:
-            self._normalize = normalize
+        self._normalize = kvargs.get('normalize', self._normalize)
         if self._normalize is True:
             self.transform = self._norm_transform
 
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -1275,8 +1275,13 @@ class Device(_Connection):
         self._nc_transform = self.transform
         self._norm_transform = lambda: JXML.normalize_xslt.encode('UTF-8')
 
-        normalize = kvargs.get('normalize', self._normalize)
-        if normalize is True:
+        # normalize argument to open() overrides normalize argument value
+        # to __init__(). Save value to self._normalize where it is used by
+        # normalizeDecorator()
+        normalize = kvargs.get('normalize')
+        if normalize is not None:
+            self._normalize = normalize
+        if self._normalize is True:
             self.transform = self._norm_transform
 
         gather_facts = kvargs.get('gather_facts', self._gather_facts)
--- a/lib/jnpr/junos/console.py
+++ b/lib/jnpr/junos/console.py
@@ -205,9 +205,7 @@ class Console(_Connection):
         # normalize argument to open() overrides normalize argument value
         # to __init__(). Save value to self._normalize where it is used by
         # normalizeDecorator()
-        normalize = kvargs.get('normalize')
-        if normalize is not None:
-            self._normalize = normalize
+        self._normalize = kvargs.get('normalize', self._normalize)
         if self._normalize is True:
             self.transform = self._norm_transform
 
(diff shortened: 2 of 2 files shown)


jnpr-community-netdev:

vnitinv:
ok to test

vnitinv:
ok to test

lib/jnpr/junos/device.py
```diff
@@ -1275,8 +1275,13 @@ def open(self, *vargs, **kvargs):
         self._nc_transform = self.transform
         self._norm_transform = lambda: JXML.normalize_xslt.encode('UTF-8')
 
-        normalize = kvargs.get('normalize', self._normalize)
-        if normalize is True:
+        # normalize argument to open() overrides normalize argument value
+        # to __init__(). Save value to self._normalize where it is used by
+        # normalizeDecorator()
+        normalize = kvargs.get('normalize')
```
vnitinv: Why not just do

        self._normalize = kvargs.get('normalize', self._normalize)
        if self._normalize is True:
            self.transform = self._norm_transform

lib/jnpr/junos/device.py
```diff
@@ -1275,8 +1275,13 @@ def open(self, *vargs, **kvargs):
...
+        if normalize is not None:
```
vnitinv: same as first comment

PR #760, opened 2017-07-18
Facts gathering support for Junos Node Slicing

stacywsmith:
[Node Slicing](https://www.juniper.net/documentation/en_US/junos/information-products/pathway-pages/junos-node-slicing/junos-node-slicing.pdf) requires some additions to PyEZ fact gathering.
Support fact gathering for both BSYS and GNFs.


jnpr-community-netdev:

PR #761, opened 2017-07-24
PyEZ fact gathering support for JDM of Junos Node Slicing.

stacywsmith:
- The JDM instance in Junos Node Slicing requires additions to support
PyEZ fact gathering.

- Also enhanced support for GNFs in Junos Node Slicing.

- Unit tests added

- PEP8 cleanup


jnpr-community-netdev:

vnitinv:
Ok to test

vnitinv:
Ok to test

vnitinv:
ok to test

vnitinv:
ok to test

vnitinv:
ok to test

lib/jnpr/junos/facts/get_software_information.py
```diff
@@ -17,6 +17,12 @@ def _get_software_information(device):
             except Exception:
                 pass
         try:
+            # JDM for Junos Node Slicing
+            return device.rpc.cli("show version all-servers", format='xml',
```
vnitinv: by this time we dont if any data in facts to do "if" check in place of try/catch.

Also we can just use below rpc in place of cli
```
<rpc>
        <get-software-information>
        <all-servers/>
        </get-software-information>
</rpc>

Comment (stacywsmith) [inline lib/jnpr/junos/facts/get_software_information.py]: No, at the time this runs, there aren't really any facts available that we could use for an if block rather than a try/catch block.

I will convert to using the suggested RPC since it does seem to work in this case. I hadn't used the RPC because `| display xml rpc` didn't show it.
```
root@jdm> show version all-servers | display xml rpc 
<rpc-reply xmlns:junos="http://xml.juniper.net/junos/17.4I0/junos">
    <rpc>
        <get-software-information>
        </get-software-information>
    </rpc>
    <cli>
        <banner></banner>
    </cli>
</rpc-reply>
```

vnitinv: " I hadn't used the RPC because | display xml rpc didn't show it." This is a long pending bug.

PR #762, opened 2017-07-25
Setting no-resolve to true for faster ARP Table lookups

rucarrol:
By default we do not set 'no-resolve' for get-arp-table-information. 

While this mimics JunOS default behaviour, it can easily tip the default 30s timeout for busy devices. 

This PR is to default set no-resolve, so we do not spend the time on the JunOS device to do DNS lookups, rather give the client the option to do this.


jnpr-community-netdev:

stacywsmith:
@rucarrol Thanks for the pull request.

PR #772, opened 2017-08-14
Softwire example

mwiget:
Adding ietf-softwire get_config() example call with a URN formatted namespace.


jnpr-community-netdev:

PR #773, opened 2017-08-16
Add vmhost parameter to SW.install() to support upgrading the VM Host.

stacywsmith:
Some Junos REs/platforms run Junos in a VM on a Linux host os. On those platforms, `junos-vmhost-* ` packages are published which upgrade both the Linux host os and the Junos VM. These `junos-vmhost-*` packages are installed with a different NETCONF RPC, `<request-vmhost-package-add>`.

Adding a new `vmhost` parameter to `SW.install()` if `vmhost=True` is specified, then the package is installed using the `<request-vmhost-package-add>` RPC. If `vmhost=False` (the default), then the package continues to be installed with the `<request-vmhost-package-add>` RPC.

### DIFF
--- a/lib/jnpr/junos/utils/sw.py
+++ b/lib/jnpr/junos/utils/sw.py
@@ -679,21 +679,21 @@ class SW(Util):
                    to reboot the device.
 
         :param str package:
-          Either the full file path to the install package tarball on the local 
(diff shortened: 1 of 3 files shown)


jnpr-community-netdev:

vnitinv:
ok to test

PR #782, opened 2017-09-24
SRX Branch cluster fails SW.install().

stacywsmith:
On SRX branch devices configured in an SRX cluster, `SW.install()`
fails because it recognizes the cluster as `'2RE' == True`, and attempts
to upgrade both REs. However, this product doesn't not support the
`re0` and `re1` arguments.

This change avoids attempting to upgrade both REs of the cluster,
even when `all_re = True`.

### DIFF
--- a/lib/jnpr/junos/utils/sw.py
+++ b/lib/jnpr/junos/utils/sw.py
@@ -972,7 +972,7 @@ class SW(Util):
         else:
             cmd = E('request-reboot', E('at', str(at)))
 
-        if all_re is True :
+        if all_re is True:
             if self._multi_RE is True and self._multi_VC is False:
                 cmd.append(E('both-routing-engines'))
             elif self._mixed_VC is True:
--- a/lib/jnpr/junos/utils/sw.py
+++ b/lib/jnpr/junos/utils/sw.py
@@ -876,7 +876,9 @@ class SW(Util):
                     "NSSU: installing software ... please be patient ...")
                 return self.pkgaddNSSU(remote_package,
                                        dev_timeout=timeout, **kwargs)
-            elif self._multi_RE is False or all_re is False:
+            elif (self._multi_RE is False or all_re is False or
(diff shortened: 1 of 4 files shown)


stacywsmith:
Addresses Juniper/ansible-junos-stdlib#261

lib/jnpr/junos/utils/sw.py
```diff
@@ -967,7 +972,7 @@ def reboot(self, in_min=0, at=None, all_re=True):
         else:
             cmd = E('request-reboot', E('at', str(at)))
 
-        if all_re is True:
+        if all_re is True :
```
vnitinv: why extra space?

lib/jnpr/junos/utils/sw.py
```diff
@@ -72,6 +72,11 @@ def __init__(self, dev):
             self._RE_list = [x for x in dev.facts.keys()
                              if x.startswith('version_RE')]
         self._multi_RE = bool(dev.facts.get('2RE'))
+        # Branch SRX in an SRX cluster doesn't really support multi_RE
+        # functionality for SW.
+        if (dev.facts.get('personality', '') == 'SRX_BRANCH' and
```
vnitinv: This change won't send re0 to rpc call, but do we need to install for different node explicitly.

lib/jnpr/junos/utils/sw.py
stacywsmith: That was an accident. Thanks for catching it. I have corrected it.

lib/jnpr/junos/utils/sw.py
stacywsmith: Unfortunately, this platform/configuration does not support the ability to upgrade a node other than the currently connected node. Therefore, it must be the user's responsibility to explicitly perform the `SW.install()` on each node of the cluster.

PR #788, opened 2017-10-21
Maintain ordering on JSON returned from the device.

stacywsmith:
The order of JSON object/dictionary keys is sometimes important in
Junos. This change ensures that order is maintained when the JSON
string is converted to a Python data structure. Because an
`OrderedDict` is an instance of a `dict`, this should be completely
backwards compatible.


stacywsmith:
Closing this for now. It has a problem on Python 2.6, and I'm also not yet sure this is the correct long-term fix.

PR #791, opened 2017-11-02
Detect format "set" when using "insert" keyword

vincentbernat:


stacywsmith:
@vincentbernat Thanks for the contribution!

PR #792, opened 2017-11-03
Detect set config format with other keywords

pklimai:
Following #791 and being consistent, we should also include activate|deactivate|annotate|copy|protect|unprotect to the regex. I tested and it works. Tell me if a unit test per keyword is needed (we currently have them for set, delete and insert).


pklimai:
Thanks for a quick review @stacywsmith - will do.

pklimai:
@stacywsmith - added unit tests. Also, I removed 'replace' (which was there initially) from regex: it turns out you can't load a config containing 'replace' keyword (either with PyEZ or with CLI), e.g.
```
lab@vMX-1# load set terminal 
[Type ^D at a new line to end input]
replace pattern 1.1.1.1 with 2.2.2.2
terminal:1:(0) unknown command: replace
load complete
```

stacywsmith:
@pklimai At first, I was wondering why you removed `replace`. I now see that it isn't allowed when loading configuration in set format.

Documenting that it isn't supported for future reference in case anyone has the same question I did:
```
<rpc><load-configuration action="set" format="text"><configuration-set>replace pattern ge-0/0/0 with ge-0/0/25</configuration-set></load-configuration></rpc>
<rpc-reply xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" xmlns:junos="http://xml.juniper.net/junos/17.3R1/junos">
<load-configuration-results>
<rpc-error>
<error-type>protocol</error-type>
<error-tag>operation-failed</error-tag>
<error-severity>error</error-severity>
<error-message>unknown command</error-message>
<error-info>
<bad-element>replace</bad-element>
</error-info>
</rpc-error>
<ok/>
</load-configuration-results>
</rpc-reply>
]]>]]>
```

PR #798, opened 2017-11-22
'or' logic doesn't allow 'None' which evaluates as False.

jjico:
This means that "allow_agent" cannot be set to True
based on the logic provided.

Instead: check if the value exists in the dict, then
use the value absolutely if set (even if it's "None").


jnpr-community-netdev:

jnpr-community-netdev:

jjico:
This failed because paramiko returned "zero length field name in format", which is an old python2.6-ism with format strings. The failure wasn't with the patch itself. Paramiko updated about 12 days ago and paramiko 2.X no longer supports python 2.6 ( https://github.com/paramiko/paramiko/issues/1070 ). I submitted a patch for this and Travis seems to have accepted it.

Separately I agree with paramiko and think it's reasonable to drop python2.6 support, but that's not mine to decide.

Please re-evaluate.

vnitinv:
@jjico Can you please explain how `or` operator is evaluating to False. `None or None` will only give `None`.

jjico:
If you have a "Host *" entry in your ~/.ssh/config, then "self._conf_ssh_private_key_file" will be set to the sshkey provided there, paramiko correctly finds it for you.

jjico:
The issue remains but won't be resolved without a massive rebase. Closing this pull request without merging.

rsmekala:
@jjico This feature request will be taken in #920. Can you please provide your comments, if you need anything more.

PR #802, opened 2018-01-10
For MX80 platform, have SW.install() recognize error when package doesn't exist. Fixes #801.

### LINKED ISSUE #801: SW._parse_pkgadd_response() does not correctly parse output from MX80 with nonexistent package.
It appears this is a Junos issue specific to the MX80 platform. If I attempt to install a package which doesn't exist, then this platform produces:
```
user@mx80> request system software add /var/tmp/nonexistent.tgz | display xml 
<rpc-reply xmlns:junos="http://xml.juniper.net/junos/17.3R1/junos">
    <cli>
        <ignore-signals>
            hup
        </ignore-signals>
    </cli>
    <output>
        ERROR: package: /var/tmp/nonexistent.tgz is not found or empty
    </output>
    <cli>
        <banner></banner>
    </cli>
</rpc-reply>
```

Notice that the error message produced is NOT an RPC error, but just a text string in an `<output>` element which happens to contain the string `ERROR`. There's also no `<package-result>` element.

On a VMX running the same version of Junos, I see:
```
user@vmx> request system software add /var/tmp/nonexistent.tgz | display xml 
<rpc-reply xmlns:junos="http://xml.juniper.net/junos/17.3R1/junos">
    <cli>
        <ignore-signals>
            hup
        </ignore-signals>
    </cli>
    <output>
        /usr/libexec/ui/package: /var/tmp/nonexistent.tgz: no such file
        ERROR: No packages added
    </output>
    <package-result>1</package-result>
    <cli>
        <banner></banner>
    </cli>
</rpc-reply>
```
This platform correctly indicates a problem by the non-zero value for the `<package-result>` element.

There are other platforms which produce an RPC error upon error, and don't include a `<package-result>` element upon success. For this reason, PyEZ considers the MX80 output shown above to indicate that the package was successfully installed.

Instead, this MX80 output shown above should cause SW._parse_pkgadd_response() to return a non-zero return code indicating an error.

stacywsmith:
On the MX80 platform. If I attempt to install a package which doesn't exist, then this platform produces:
```
user@mx80> request system software add /var/tmp/nonexistent.tgz | display xml
<rpc-reply xmlns:junos="http://xml.juniper.net/junos/17.3R1/junos">
    <cli>
        <ignore-signals>
            hup
        </ignore-signals>
    </cli>
    <output>
        ERROR: package: /var/tmp/nonexistent.tgz is not found or empty
    </output>
    <cli>
        <banner></banner>
    </cli>
</rpc-reply>
```
Notice that the error message produced is NOT an RPC error, but just a text string in an `<output>` element which happens to contain the string ERROR. There's also no `<package-result>` element.

There are also other known cases where a missing `<package-result>` element indicates the package was installed successfully.

This change checks for this specific error message by checking if the string `'ERROR: package:'` and the string `'is not found or empty'` are present in the `<output>` tag(s).


jnpr-community-netdev:

vnitinv:
ok to test

jnpr-community-netdev:
Build triggered. sha1 is merged.

jnpr-community-netdev:
Build started sha1 is merged.

jnpr-community-netdev:
Build finished.

rahkumar651991 (approved):

PR #821, opened 2018-04-09
Created op views for dns-proxy-information & system-memory

pr11t:


jnpr-community-netdev:

vnitinv:
@pr11t  Any specific reason why you opted for all fields as key?

pr11t:
@vnitinv  I just figured this won't work with a chassis cluster ... Any tips on making it work without creating a separate files? Otherwise might have to scratch this patch I guess.

pr11t:
After changing the item key I tested with both cluster and not cluster config and seems to work this way

PR #827, opened 2018-05-08
Updated to use Python3

ntwrkguru:
Updated to use Python 3.6
Updated .dockerignore to clean up the build context
Added myself to the list of contributors...shamelessly 🥇


jnpr-community-netdev:

PR #828, opened 2018-05-08
Fix for broken Dockerfile

ntwrkguru:
The revised .dockerignore prevented requirements.txt from being copied to the container. This fixes that regression.


jnpr-community-netdev:

PR #848, opened 2018-07-14
Fix for XML tag being recognized as cli prompt.

jnpr-bowen:
The 'cli' pattern in the _RE_PAT is matching an xml tag like `<solo-tag/>` as if it is the cli prompt host_name>. This has caused the login state machine to fail on console connections when connecting back to a console where an RPC call has initiated a reboot.

The following shows the change in behavior from a python shell session.
Current code:
`>>> _re_pat_login = '(?P<login>ogin:\s*$)'\s`
`>>> _RE_PAT = [ `
`...    '(?P<loader>oader>\s*$)',`
`...    _re_pat_login,`
`...    '(?P<passwd>assword:\s*$)',`
`...    '(?P<badpasswd>ogin incorrect)',`
`...    '(?P<netconf_closed><!-- session end at .*-->\s*)',`
`...    '(?P<shell>%|#\s*$)',`
`...    '(?P<cli>[^\\-"]>\s*$)',`
`...    '(?P<option>Enter your option:\s*$)',`
`...    '(?P<hotkey>connection: <CTRL>Z)',`
`... ]`
`>>> _PROMPT = re.compile(six.b('|').join([six.b(i) for i in _RE_PAT]))`
`>>> test_str = "<a_tag> 'Tag text'</a_tag> <solo_tag/>"`
`>>> found = _PROMPT.search(test_str)`
`>>> print found.lastgroup`
cli

New code changing the 'cli' pattern:

`>>> _RE_PAT = [`
`...     '(?P<loader>oader>\s*$)',`
`...     _re_pat_login,`
`...     '(?P<passwd>assword:\s*$)',`
`...     '(?P<badpasswd>ogin incorrect)',`
`...     '(?P<netconf_closed><!-- session end at .*-->\s*)',`
`...     '(?P<shell>%|#\s*$)',`
`...     '(?P<cli>[^/\-"]>\s*$)',`
`...     '(?P<option>Enter your option:\s*$)',`
`...     '(?P<hotkey>connection: <CTRL>Z)',`
`... ]`
`>>> _PROMPT = re.compile(six.b('|').join([six.b(i) for i in _RE_PAT]))`
`>>> test_str = "<a_tag> 'Tag text'</a_tag> <solo_tag/>"`
`>>> found = _PROMPT.search(test_str)`
`>>> print found.lastgroup`
  File "<input>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'lastgroup'

Which is correct behavior, the tag is not recognized as the cli prompt.

### DIFF
--- a/lib/jnpr/junos/transport/tty.py
+++ b/lib/jnpr/junos/transport/tty.py
@@ -271,5 +271,3 @@ class Terminal(object):
         else:
             # if we are here, then loop the event again
             self._login_state_machine(attempt + 1)
-
-        _ev_tbl.get()
--- a/lib/jnpr/junos/transport/tty.py
+++ b/lib/jnpr/junos/transport/tty.py
@@ -52,7 +52,7 @@ class Terminal(object):
         '(?P<badpasswd>ogin incorrect)',
         '(?P<netconf_closed><!-- session end at .*-->\s*)',
         '(?P<shell>%|#\s*$)',
-        '(?P<cli>[^\\-"]>\s*$)',
+        '(?P<cli>[^/\-"]>\s*$)',
         '(?P<option>Enter your option:\s*$)',
         '(?P<hotkey>connection: <CTRL>Z)',
     ]
@@ -271,3 +271,5 @@ class Terminal(object):
         else:
             # if we are here, then loop the event again
             self._login_state_machine(attempt + 1)
+
+        _ev_tbl.get()


jnpr-community-netdev:

vnitinv:
@jnpr-bowen What if CLI prompt is something like 
`user@xyz/> `

jnpr-bowen:
That will not be recognized as the cli prompt, just as the following are not with the current code:
user@xyz->
user@xyz”>
Both of these are attempts to match xml tags like <!—comment -->, and <tag=”some text”>. The one the current code is not avoiding is the solo tag <solo/>. The only one of the characters (-, “, and /) that can show up in a host name is -, so a host named foo- would create a failure to recognize the cli prompt, but a host name cannot include “, or /. The only way to get those to show up in the cli prompt would be to use the “set cli prompt” command that does not appear last beyond the current cli session.
This change should be safe, and if need be, some documentation in pyez could say to not end host names with a -.

Later,
Bruce

jnpr-community-netdev:

jnpr-community-netdev:

rahkumar651991:
this is a corner case scenario. Adding it will create issue for other scenario. 
Hence closing this.

PR #850, opened 2018-07-16
Checking for the presence of xml in err

bcavns01:
Checking for the presence of xml in err before attempting to use it for composing an rpc_error.


jnpr-community-netdev:

Issue #864, opened 2018-08-18
sw.install() not catching multiple <package-result> values


jtishey:
The following example returns "True" on an EX4200 when install fails due to corrupt image file on the device.

```
from jnpr.junos.utils.sw import SW
ok = SW(self.dev).install(package=PACKAGE, validate=False, reboot=True, no_copy=True,
                        progress=True, remote_path=R_PATH)
```

It appears that a package-result of "0" is getting returned on the first response, followed by a second value of "7" when the install fails, but only the first value is getting matched, which makes it look like there was no error.

```
>>> r = dev.rpc.request_package_add(reboot=True, no_validate=True, package_name=PACKAGE)
>>> r
<Element pipe at 0x7f2771beae88>
>>> got = r.getparent()
>>> got.findtext('package-result')
'0'
>>> from lxml import etree
>>> 
>>> etree.tostring(got)
b'<rpc-reply message-id="urn:uuid:72b10d20-f14d-48b0-8eb6-7c3a41cb3cf9">\n<pipe>\n<more-no-more/>\n</pipe>\n<cli>\n<ignore-signals>\nhup\n</ignore-signals>\n</cli>\n<output>\n\nChecking pending install on fpc0</output>\n<output/>\n<package-result>0</package-result>\n<output>\n\nfpc0:</output>\n<pipe><more-no-more/>\n</pipe>\n<cli>\n<ignore-signals>hup</ignore-signals>\n</cli>\n<output>\ngunzip: /var/tmp/jinstall-ex-4200-12.3R6.6-domestic-signed.tgz: unexpected end of file\ntruncated or corrupted package: /var/tmp/jinstall-ex-4200-12.3R6.6-domestic-signed.tgz</output>\n<package-result>7</package-result>\n</rpc-reply>'
```

The install log on the switch shows the following:
```
*** install ***
2018-04-27 23:19:41 UTC mgd[2867]: /usr/libexec/ui/package -X check-pending-install
<output>
</output>
<package-result>0</package-result>
2018-04-27 23:19:41 UTC mgd[2867]: /usr/libexec/ui/package -X update -no-validate /var/tmp//jinstall-ex-4200-12.3R6.6-domestic-signed.tgz
<output>

gunzip: /var/tmp//jinstall-ex-4200-12.3R6.6-domestic-signed.tgz: unexpected end of file
truncated or corrupted package: /var/tmp//jinstall-ex-4200-12.3R6.6-domestic-signed.tgz
</output>
[...]

PR #866, opened 2018-08-22
Look for multiple package-result values from sw.install #864

jtishey:
If multiple <package-result> values are returned, only the first one is being matched, while additional values are ignored.  This can indicate success on failed installs.

This request is to parse multiple values and return False (failed) if any are non-zero.


rsmekala:
@jtishey Can you please add UT for the code before we merge it ??

 Tests to be added in: https://github.com/Juniper/py-junos-eznc/blob/master/tests/unit/utils/test_sw.py

rsmekala:
Addresses #864

vnitinv:
@rsmekala Can we will only add UT for the code change. Please take this up. Thanks

PR #868, opened 2018-08-22
Add escaped $ Bourne-style shell prompt support

mmellin:
This pull request adds proper support by escaping the `$` character within the shell prompt variables.  This avoids errors when running under normal CLI prompt conditions.


jnpr-community-netdev:

mmellin:
I guess I didn't create a new branch to split these commit into.  Sorry.

Commit `fcecf2e` is an update to fix an issue with adding the `$` prompt.  This character seems to need escaping in order for the code to work properly under all conditions.

Commit `a9a9c57` is in response to issue #869

jnpr-community-netdev:

dineshbaburam91:
Fixed using #882 and #934.
Kindly let us know why you required "sleep" in wait_for() definition?

ntwrkguru:
@dineshbaburam91 The reason for the sleep was literally in the code:

```
        :param seconds sleep:
           Time to wait after initial call to receive data from buffer. This
           value can help stabilize the output when multiple calls to run()
           are looped but will increase the time spent receiving output.  This
           value can be a floating point number for subsecond precision.
```

rahkumar651991 (approved):

PR #870, opened 2018-08-27
Console ssh via device class through Device class

vnitinv:
Example for usage:

```python
from jnpr.junos import Device
import logging

logging.basicConfig(level=logging.DEBUG)

with Device(host="xx.xx.xx.xx", user="xxx", passwd="xxx", cs_user='xxxx', cs_passwd='xxx') as dev:
    print(dev.cli("show version"))
```


lboue:
Thanks for this commit. 
Could you be more specific about console server support? Which device has been tested?

Regards

Issue #877, opened 2018-09-06
Console over SSH not working in 2.2.0 


dgarros:
@vnitinv @mzbroch 
The SSH over console feature introduced in 2.2.0 is not working for me
I wanted to do the integration with the juniper_junos ansible module but it looks broken in Pyez 2.2.0

I did a bit of troubleshooting and I was able to make it work when I change the flag `RECVSZ` to 1, current value is 1024 (in file `lib/jnpr/junos/transport/tty_ssh.py`)

@vnitinv, looks like you changed it from 1 to 1024 in #870, any specific raison for that ? 

Happy to send a PR to flip it back to 1 if we agreed it's the right fix

mzbroch:
@vnitinv @dgarros 

RECVSZ has been set to minimum value of 1 to chunk data and allow for reading newline characters. Read, readline, read_until are method commonly used for other transports already implemented and I took a design decision to mimic same behaviour for ssh transport. 

According to my tests, setting higher value of RECVSZ will break the feature in *most* cases, this is due to standard behaviour that declared portion of the SSH (Paramiko) buffer data will be read, however it can be more data than matched/expected by the regexp.

As there are already many sleeps (both time.sleep and select) within the code in multiple places, setting value of 1 to RECVSZ as per original commit had no significant impact on the performance during tests performed - my user experience during transfers with latency ~150ms was very good in this case, as the lowest buffer is not being used for the read_until method which generates the most traffic (it transfers netconf data).

dgarros:
Thanks @mzbroch for these information, I created a PR to change this value
#881

vnitinv:
@dgarros  Reason for changing it to 1024, Over SSH do we really need such slow read.
with setting it to 1, the state machine will be called on every char read. Will impact performance.

On CLI we sometimes see a message like
`Type the hot key to suspend the connection: <CTRL>Z`
with every char read, when it reaches `Type the hot key to suspend the connection: <CTRL>`
(without Z) it match CLI option and hence state machine breaks down to login.
https://github.com/Juniper/py-junos-eznc/blob/master/lib/jnpr/junos/transport/tty.py#L55

We need to see why 1024, is cauing issue on your setup and fix accordingly.

@stacywsmith any comment?

dgarros:
@vnitinv, have you been able to test the code successfully with 1024 ? based on @mzbroch comment, it looks like it shouldn't work in most cases.

[...]

PR #881, opened 2018-09-11
Change Receive Size to 1 for transport_ssh

dgarros:
Fix for #877 
Based on @mzbroch, received size must be 1 for this library to work


jnpr-community-netdev:

vnitinv:
@dgarros This will not only slow down login but also subsequent steps, for ex: RPC calls

jnpr-community-netdev:

vnitinv:
We already now have tested code for login via ssh console server. Hence closing this pull request.

PR #896, opened 2018-11-04
create generalized function for new ssh-client

jeremyschulman:
open_ssh_client(device) was created so that a developer can create a new paramiko SSH client based on an existing device credentials.  By providing this function, the developer can then create any new connections to the Junos device, for example an SCP client or a SFTP client. 

This PR includes refactoring the existing scp module to use this new function.

### DIFF
--- a/lib/jnpr/junos/utils/scp.py
+++ b/lib/jnpr/junos/utils/scp.py
@@ -3,6 +3,7 @@ import inspect
 
 import paramiko
 from scp import SCPClient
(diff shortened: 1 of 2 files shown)


vnitinv:
@jeremyschulman Can you also check if we can also make use of this API in StartShell class?

jnpr-community-netdev:

jnpr-community-netdev:

vnitinv:
Taken care via #957

PR #904, opened 2018-12-18
Update README.md

webberist:
The provided docker example causes an error when mapping paths with spaces. Using -v "$(pwd)":/scripts is more correct as per https://docs.docker.com/storage/bind-mounts/


jnpr-community-netdev:

jnpr-community-netdev:

PR #905, opened 2019-01-10
add at parameter to SW.poweroff()

shigechika:
SW.reboot() supports **at** parameter. but SW.poweroff() does not support.
This patch add **at** parameter to SW.poweroff().
Thank you.

### DIFF
--- a/lib/jnpr/junos/utils/sw.py
+++ b/lib/jnpr/junos/utils/sw.py
@@ -1005,7 +1005,7 @@ class SW(Util):
     # poweroff - system shutdown
     # -------------------------------------------------------------------------
 
-    def poweroff(self, in_min=0, on_node=None):
+    def poweroff(self, in_min=0, at=None, on_node=None):
         """
         Perform a system shutdown, with optional delay (in minutes) .
 
@@ -1013,6 +1013,9 @@ class SW(Util):
         rebooted.  This code also handles EX/QFX VC.
 
         :param int in_min: time (minutes) before shutting down the device.
+
+        :param str at: date and time the poweroff should take place. The
+            string must match the junos cli poweroff syntax
 
         :param str on_node: In case of linux based device, function will by default
             shutdown the whole device. If any specific node is mentioned,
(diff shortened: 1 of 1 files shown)


jnpr-community-netdev:

shigechika:
This patch works fine on my environment.
Please verify.

lib/jnpr/junos/utils/sw.py
```diff
@@ -1037,7 +1040,10 @@ def poweroff(self, in_min=0, on_node=None):
             cmd = E('request-power-off')
             if self._multi_RE is True and self._multi_VC is False:
                 cmd.append(E('both-routing-engines'))
-        cmd.append(E('in', str(in_min)))
+        if in_min >= 0 and at is None:
```
rsmekala: Order of preference in JUNOS: `in_min` is honored before `at`.  Can you change the if condition to do the same ??

For reference, logic in Ansible:
https://github.com/rsmekala/ansible-junos-stdlib/blob/9fa1ddc899923dcc224d9cce272ae55d9fc20966/library/juniper_junos_system.py#L377

shigechika: Yes, this patch works as your expects.
* 1st priority: in_min
* 2nd priority: at
Thank you.
## This patch referenced SW.reboot().

rsmekala: The expected behavior is not met in the current logic.

On executing the below script:
```
from jnpr.junos.device import Device
from jnpr.junos.utils.sw import SW

with Device('xx.xx.xx.xx', user='xxx', passwd='xxxxx') as dev:
    sw = SW(dev)
    sw.poweroff(in_min=2, at="1903182100")
```

I see the below netconf trace.
```
Mar 18 20:01:59 [NETCONF] - [12475] Incoming: <?xml version="1.0" encoding="UTF-8"?><nc:rpc xmlns:nc="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="urn:uuid:30b716f5-26e2-4cbb-9cc9-2258f942bdba"><request-power-off><bot
h-routing-engines/><at>1903182100</at></request-power-off></nc:rpc>]]>]]>
```

The current logic checks if `in_min >= 0 and at is None` which is `False` in the above case hence `at` is appended instead of `in_min`

rsmekala: Thanks for pointing out `sw.reboot`, will fix it in another PR.

shigechika:
Thank you for your review.
I understand my bug. I’ll fix soon.

2019年3月18日(月) 23:41 rsmekala <notifications@github.com>:

lib/jnpr/junos/utils/sw.py
shigechika: Thank you for your review.
I understand my bug. I’ll fix it soon in another PR.

PR #911, opened 2019-02-13
Replaced "python setup.py install" with "pip3 install" in Dockerfile

carpalex:
Because `junos-eznc` is a namespaced package, installing it with `python setup.py install` does not correctly generate `.pth` files in `/var/lib/python3.6/site-packages/`.
If other packages that belong to the `jnpr` package are installed with `pip` (which is almost always the case), the namespace mechanism breaks and imports fail.
Replacing `python setup.py install` with `pip3 install .` fixes this issue.


carpalex:
@vnitinv @ntwrkguru can you take a look?
We need this fix mainly for using the image along with other packages in the `jnpr` namespace, as part of J-EDI development in PS Automation.

PR #916, opened 2019-03-19
fix reboot and poweroff behavior. support poweroff(at)

shigechika:
I fixed reboot() and poweroff() behavior:
1. in_min
1. at

This logic is same as the ansible.
And added supports poweroff(at).
@rsmekala, Thank you for your advices.

Thank you.


jnpr-community-netdev:

lib/jnpr/junos/utils/sw.py
```diff
@@ -986,7 +986,7 @@ def reboot(self, in_min=0, at=None, all_re=True, on_node=None):
                     cmd.append(E('both-routing-engines'))
                 elif self._mixed_VC is True:
                     cmd.append(E('all-members'))
-        if in_min >= 0 and at is None:
+        if in_min >= 0:
```
rsmekala: The default value of `in_min` is 0. So this condition is always true.

The test case is failing due to this.

A better logic would be 
```
# in_min given
if in_min > 0:
    append in

# in_min not given but at given
elif at is not None:
    append at

# default case
else:
    apppend in = 0
```

shigechika: It's good idea!

shigechika:
Please verify again.

jnpr-community-netdev:

shigechika:
I am using this patch since several months ago.
Of course, works fine.
-- Shigechika AIKAWA

vnitinv:
Sorry for getting late on this @shigechika 
How about changing this

```
        if in_min > 0:
            # in_min given
            cmd.append(E('in', str(in_min)))
        elif at is not None:
            # in_min not given but at given
            cmd.append(E('at', str(at)))
        else:
            # default case (in_min=0)
            cmd.append(E('in', str(in_min)))
```
to

```
        if in_min >= 0:
            # in_min given or default case
            cmd.append(E('in', str(in_min)))
        elif at is not None:
            # in_min not given but at given
            cmd.append(E('at', str(at)))
```

rsmekala:
@shigechika 'at' and 'in' should be mutually exclusive options in Junos. Tracked via PR# 1461215

rsmekala:
ok to test

rsmekala:
@shigechika Thanks for the contribution. Merging the pull request. ✌️

PR #928, opened 2019-04-15
Fix #927 (Cannot set Device.open timeout)

### LINKED ISSUE #927: Cannot set Device.open timeout
`jnpr/junos/device.py Device.open` does not allow passing a `timeout` parameter that would properly propagate all the way down to `ncclient/transport.ssh.py:connect`. A possible workaround can be implemented in ncclient (https://github.com/ncclient/ncclient/pull/312),  but it's not a proper fix.

When calling open on a Device that is unreachable, ssh will timeout after ~2 minutes (Linux platform). This is not an acceptable working condition when managing a network that has hundreds of Juniper devices.

Also, the following documentation for `jnpr/junos/device.py Device.open(self, *vargs, **kwargs)`:
is wrong:
```
        :raises ConnectTimeoutError:
            When the the :meth:`Device.timeout` value is exceeded
            during the attempt to connect to the remote device
```

This exception is raised when the ncclient cannot perform the ssh connection and the Device.timeout has no effect on this. Device.timeout can only be set once a valid Device._conn exists, and says how long it should wait for an aswer on an already _established_ connection.

sstancu:
Fixes #927

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -997,6 +997,7 @@ class Device(_Connection):
             dev.open()   # this will probe before attempting NETCONF connect
 
     """
+    _MIN_CONNECT_TIMEOUT = 3  # seconds
 
     # -------------------------------------------------------------------------
     # PROPERTIES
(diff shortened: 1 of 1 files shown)


jnpr-community-netdev:

sstancu:
Closing as it's not working as expected.
`ncclient.manager.connect_ssh` passes its `**kwds` in two places:
* to `transport.SSHSession` (here timeout -> connection timeout)
* to `Manager` (here timout -> rpc request timeout

Therefore without a ncclient enhancement this feature can't be achieved.

PR #932, opened 2019-04-23
added callback functionality to ftp get

ipmonk:


jnpr-community-netdev:

lib/jnpr/junos/utils/ftp.py
```diff
@@ -97,13 +97,20 @@ def get(self, remote_file, local_path=os.getcwd()):
...
-                            open(local_file, 'wb').write)
-        except Exception as ex:
-            logger.error(ex)
-            return False
-        return True
+
+        with open(local_file, 'wb') as local_fh:
+            def callback(data):
+                local_fh.write(data)
+                args_callback = self._ftpargs.get('callback')
```
vnitinv: this line should be moved above def callback, unnecessarily we will call it with each chunk of data received.

PR #934, opened 2019-04-30
improve StartShell support for Bourne shell

ipmonk:
two things in this merge request
1) add Bourne shell prompt to wait_for to prevent unnecessary wait for cmd to return
2) if hostname contains a '0', and the shell prints the hostname as part of StartShell.run() rc, then last_ok will always be True. 

Please consider merging this patch, thanks!


jnpr-community-netdev:

stacywsmith:
Looks like a good fix to me.

PR #939, opened 2019-05-21
Adding yaml for els style get-ethernet-switching-table-information rpc

ipmonk:


jnpr-community-netdev:

PR #952, opened 2019-07-29
support for vmhost reboot

### LINKED ISSUE #948: reboot vmhost
sw.reboot should also support for junos platform with vmhost.

vnitinv:
Fix #948


lib/jnpr/junos/utils/sw.py
```diff
@@ -945,8 +946,9 @@ def _progress(report):
     # -------------------------------------------------------------------------
     # reboot - system reboot
     # -------------------------------------------------------------------------
-
-    def reboot(self, in_min=0, at=None, all_re=True, on_node=None):
+    @timeoutDecorator
+    def reboot(self, in_min=0, at=None, all_re=True, on_node=None,
+               vmhost=False, **kwargs):
```
rsmekala: Why is `**kwargs` added, where is it used in the function ??

lib/jnpr/junos/utils/sw.py
```diff
@@ -979,29 +985,40 @@ def reboot(self, in_min=0, at=None, all_re=True, on_node=None):
...
                 got = rsp.getparent().findtext('.//request-reboot-status').strip()
+                if got is None:
+                    # Oon some platforms stopping/rebooting both
+                    # REs produces <output> messages and
+                    # <request-reboot-status> messages.
+                    output_msg = '\n'.join([i.text for i in got.findall('output')
+                                            if i.text is not None])
+                    if output_msg is not '':
+                        got = output_msg
             return got
```
rsmekala: Will `got` be equal to `None` in case Junos device does not output anything ??

Can we change it, or should we continue with `None` to be backward-compatible ??

lib/jnpr/junos/utils/sw.py
vnitinv: Fixed via https://github.com/Juniper/py-junos-eznc/pull/952/commits/fc6cf16a7c66a5ac0d950b920e59bf698a51295e

lib/jnpr/junos/utils/sw.py
vnitinv: I have removed split as can be an issue when findtext returns None. https://github.com/Juniper/py-junos-eznc/pull/952/commits/fc6cf16a7c66a5ac0d950b920e59bf698a51295e#diff-d1aa66cff516a1461b58eda901c5a52dR1007

got should have some text as we are handling None case where we check for  `<output>`

PR #956, opened 2019-09-24
handle error with exit for context manager

vnitinv:
In case of close throwing any exception, the context manager was erroring out. This should not be the case while doing exit.


vnitinv:
ok to test

vnitinv:
ok to test

vnitinv:
ok to test

rsmekala:
retest this please

PR #971, opened 2019-10-24
option to ignore warnings in config diff

odeshog:
Some junos devices gives warnings when doing a diff of config. Probably a problem with how Junos reports it, but it doesn't hurt to be able to ignore warnings when doing a comparison.

Example of problem device and config:
```
root@QFX5120# show | compare               
[edit chassis fpc 0 pic 0 port 0 speed]
  'speed 1G'
    warning: 1g config will be applied to ports 0 to 3
[edit]
-  chassis {
-      fpc 0 {
-          pic 0 {
-              port 0 {
-                  speed 1G;
-              }
-          }
-      }
-  }
```

### DIFF
--- a/lib/jnpr/junos/utils/config.py
+++ b/lib/jnpr/junos/utils/config.py
@@ -208,7 +208,7 @@ class Config(Util):
     # show | compare rollback <number|0*>
     # -------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


jnpr-community-netdev:

PR #998, opened 2020-03-10
add the option to specify timeout for commit_check()

dstamatov:
commit_check() does not allow to set a different timeout for the rpc other than the default 30s. For some devices this interval is not sufficient and sometimes causes otherwise valid commit check to fail.

### DIFF
--- a/lib/jnpr/junos/utils/config.py
+++ b/lib/jnpr/junos/utils/config.py
@@ -172,11 +172,13 @@ class Config(Util):
     # commit check
     # -------------------------------------------------------------------------
 
-    def commit_check(self):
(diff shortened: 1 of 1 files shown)


jnpr-community-netdev:

vnitinv:
Just for future reference

We want timeout things to be in sync everywhere. Will be dev_timeout for everything. Right now its dev_timeout for RPC calls and timeout for commit.
We also wanted to rename/link it to rpc_timeout and eventually remove dev_timeout as dev-timeout sounds something like dev connection timeout and not RPC timeout.

Watch out for code update coming in few days or so.

PR #999, opened 2020-03-13
Fix documentation typo

zihzihtw:
I found a wrong word in the documentation of poweroff function.
If the device is equipped with dual-RE, then both RE will be "shut down". Not "rebooted".


jnpr-community-netdev:

Issue #1006, opened 2020-03-22
PyEZ overwirtes host with hostname 


a-v-popov:
If we have config like this
```
Host mx480
  HostName 192.168.100.22
  ProxyCommand ssh -W %h:%p -q lab
```
PyEZ will overwrite Host with Hostname here:
https://github.com/Juniper/py-junos-eznc/blob/6f02b26cdca98f9153ac3c83d98882b872477f23/lib/jnpr/junos/device.py#L431
It is then passed to ncclient as host. As result ncclient will be trying to get config for wrong Host and fail to get ProxyCommand.

_Originally posted by @a-v-popov in https://github.com/Juniper/py-junos-eznc/issues/153#issuecomment-601967611_

vnitinv:
@a-v-popov Alex, please check if anything extra need to be done to fix this issue in https://github.com/Juniper/py-junos-eznc/pull/1007

PR #1007, opened 2020-03-24
hostname should not be overwritten, ncclient/paramiko to take care

vnitinv:
Fix: https://github.com/Juniper/py-junos-eznc/issues/1006

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,16 +3,12 @@ matrix:
     include:
         - python: 2.7
           dist: trusty
-          sudo: false
         - python: 3.5
           dist: trusty
-          sudo: false
         - python: 3.6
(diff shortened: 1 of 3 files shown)


a-v-popov:
It seems open_ssh_client() is not going through ncclient.
https://github.com/Juniper/py-junos-eznc/blob/290decca4bcae9f74ca197afbde5f57b63f28bb6/lib/jnpr/junos/utils/ssh_client.py

At the very least hosname should be picked up after config.lookup() similar to 
https://github.com/ncclient/ncclient/blob/6a2b112388142362c2cdc43e87586e185157d038/ncclient/transport/ssh.py#L248
to maintain the current level of functionality.

PR #1016, opened 2020-04-08
ensures last_ok is correct when passing in custom 'this' to StartShell.run()

ipmonk:
use case was remote shell was tcsh, with a prompt ending in '> '. 

By passing in 'this = "(%|#|\$|\>)\s"' to StartShell.run(), the prompt is matched but, last_ok is not reflecting the cmd exit code. For example, this command returns an exit code of 1, yet ss.last_ok is True:
(True, 'cat /var/tmp/foo\r\r\ncat: /var/tmp/foo: No such file or directory\r\n\x1b[4mfoohost\x1b[m:\x1b[1m~\x1b[0m> ')


ipmonk:
close it. breaks use case where user has entered junos cli interactively

PR #1019, opened 2020-04-13
textfsm window version check

vnitinv:
* latest textfsm is not supported on windows.
* Enable travis on windows.

Right now only around 800 test cases get executed for windows and others are skipped. will fix skipped one in future.

```
Ran 796 tests in 18.407s
OK (SKIP=64)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,6 +5,9 @@ import sys
 req_lines = [line.strip() for line in open(
     'requirements.txt').readlines()]
 install_reqs = list(filter(None, req_lines))
+
+# refer: https://github.com/Juniper/py-junos-eznc/issues/1015
+# should be removed when textfsm releases >=1.1.1
 if sys.platform == 'win32':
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -5,8 +5,8 @@
 req_lines = [line.strip() for line in open(
     'requirements.txt').readlines()]
 install_reqs = list(filter(None, req_lines))
-if sys.version_info[:2] == (2, 6):
-    install_reqs.append('importlib>=1.0.3')
+if sys.platform == 'win32':
+    install_reqs.append('textfsm==0.4.1')
```
dineshbaburam91, line 9: Why did we remove **importlib** ? 

Kindly add the comment saying win32 supports **testfsm 0.4.1.**

vnitinv: We dont support python 2.6 and importlib was explicitly needed for that version.
>=2.7 python got importlib by default

PR #1031, opened 2020-05-08
changes for ansible actions of reboot, shutdown to be handled in PyEZ

rahkumar651991:
Changes done to call PyEz functions to handle reboot, shutdown and similar functionalities in junos-ansible.


lib/jnpr/junos/utils/sw.py
```diff
@@ -1028,51 +1027,41 @@ def _progress(report):
...
+    def _system_operation(
+        self, cmd, in_min=0, at=None, all_re=True, other_re=False, vmhost=False
+    ):
         """
-        Perform a system reboot, with optional delay (in minutes) or at
-        a specified date and time.
-
-        If the device is equipped with dual-RE, then both RE will be
-        rebooted.  This code also handles EX/QFX VC.
+        Send the rpc for actions like shutdown, reboot, halt  with optional
```
dineshbaburam91, line 1034: @rahkumar651991 Please run sphinx-build since you added and modified comments

lib/jnpr/junos/utils/sw.py
```diff
@@ -1099,6 +1088,57 @@ def reboot(self, in_min=0, at=None, all_re=True, on_node=None, vmhost=False):
...
+            else:
+                cmd = E("request-node-reboot")
+                cmd.append(E("node", on_node))
+        elif vmhost is True:
+            cmd = E("request-vmhost-reboot")
+        else:
+            cmd = E("request-reboot")
+
+        try:
+            return self._system_operation(cmd, in_min, at, all_re, other_re, vmhost)
```
dineshbaburam91, line 1141: Kindly follow the same parameter  order which you followed in reboot call line no 1098

self, in_min=0, at=None, all_re=True, on_node=None, vmhost=False, other_re=False

tests/unit/utils/test_sw.py
```diff
@@ -888,6 +888,25 @@ def test_sw_poweroff_multi_re_vc(self, mock_execute):
...
+        mock_execute.side_effect = RpcError(rsp=rsp)
+        self.assertRaises(Exception, self.sw.halt)
+
+    @patch("jnpr.junos.Device.execute")
+    def test_sw_halt_multi_re_vc(self, mock_execute):
+        mock_execute.side_effect = self._mock_manager
+        self.sw._multi_RE = True
+        self.sw._multi_VC = False
+        self.assertTrue("Shutdown NOW" in self.sw.halt())
+
```
dineshbaburam91, line 909: Update Zeroize testcases

lib/jnpr/junos/utils/sw.py
rahkumar651991: All the new parameter are added at the end, otherwise people having current scripts will get wrong output when they execute.

PR #1038, opened 2020-05-29
fix package requirements for windows

p-sherratt:
#1019 no longer addresses issue of textfsm import in windows because of a change in ntc-templates==1.4.2  :

ImportError: cannot import name 'clitable' from 'textfsm' (<redacted>\lib\site-packages\textfsm\__init__.py)


vnitinv:
Fixed via https://github.com/Juniper/py-junos-eznc/pull/1042

PR #1041, opened 2020-06-04
Python 3.5 is required

jessicagarrison:
Changing Installation Python version from 3.4 to 3.5 per discussion with Juniper Eng.


ntwrkguru:
LGTM

PR #1068, opened 2020-08-12
Removing namespace from output tag if in first child element

rahkumar651991:


rahkumar651991:
Fixes https://github.com/Juniper/ansible-junos-stdlib/issues/517

PR #1074, opened 2020-09-06
Used Table specific named-key string to check more specific values

ngadre-juniper:
Instead of using key as the "key" to get user inputs,
now use the last xml element itself as the "key".
This would make it more intuitive for user to fetch
based on table specific keys.


ngadre-juniper:
@rahkumar651991 @vnitinv Hi, can you please take a look

vnitinv:
@ngadre-juniper  can you explain with an example

PR #1085, opened 2020-10-14
Textfsm to be manually installed by the user

rahkumar651991:


ntwrkguru:
So, was this optional before? I never understood the requirement for ntc_templates. *If* textfsm is needed, I think it can be installed without ntc_templates. What was the use case for its inclusion originally, @rahkumar651991?

vnitinv:
@ntwrkguru we need ntc_templates. Had to add textfsm (0.4.1) very specifically for windows due to some bug.
These modules are very limited in use and were added mainly for healthbot use case. But this is causing the issue of upgrading PyEZ on-box Junos due different vendor contents. So we are making this module's installation optional and to be done only if some want to use PyEZ table/view with other vendors cli o/p contents.

ntwrkguru:
> @ntwrkguru we need ntc_templates. Had to add textfsm (0.4.1) very specifically for windows due to some bug.
> These modules are very limited in use and were added mainly for healthbot use case. But this is causing the issue of upgrading PyEZ on-box Junos due different vendor contents. So we are making this module's installation optional and to be done only if some want to use PyEZ table/view with other vendors cli o/p contents.

Thanks for the clarification @vnitinv I think this is the right decision. Healthbot should not push additional external libs to PyEZ, rather use PyEZ as the base image from which to built their use case. This is pattern and makes sense to me.

vnitinv (approved):
Looks good

PR #1088, opened 2020-10-28
support use-fast-diff in diff/pdff function

vnitinv:
```
<get-configuration compare=”rollback” use-fast-diff=”yes”/>.
It works only for rollback 0.
```


rahkumar651991 (approved):
looks fine

Issue #1089, opened 2020-11-12
Deprecation warninig - collections import for python 3.3+, stop working on 3.9


qaxi:
On python 3.8 there s deprecation warning
```
/usr/local/lib/python3.8/dist-packages/jnpr/junos/factcache.py:10
  /usr/local/lib/python3.8/dist-packages/jnpr/junos/factcache.py:10: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    class _FactCache(collections.MutableMapping):

```

tirkarthi:
This looks like a duplicate of https://github.com/Juniper/py-junos-eznc/issues/1033

rahkumar651991:
closing as it is duplicate of #1033

PR #1090, opened 2020-11-12
Inorrect import from 'collections' will stop working in Python 3.9

qaxi:
Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
Solving #1089

### DIFF
--- a/lib/jnpr/junos/factcache.py
+++ b/lib/jnpr/junos/factcache.py
@@ -1,4 +1,4 @@
-import collections
+import collections.abc
 import warnings
 from pprint import pformat
 
@@ -7,7 +7,7 @@ from jnpr.junos.facts import __doc__ as facts_doc
 import jnpr.junos.exception
 
 
-class _FactCache(collections.MutableMapping):
+class _FactCache(collections.abc.MutableMapping):
     """
     A dictionary-like object which performs on-demand fact gathering.
 


qaxi:
The only fail in Travis is on Python 2.7 tests ...

Python 2.7 is deprecated from 1. January 2020 ... it is dangerous to use it.

ntwrkguru:
> Python 2.7 is deprecated from 1. January 2020 ... it is dangerous to use it.

Why is it "dangerous"? 

Less argumentatively though, on-box Python is still 2.7 in many supported Junos releases, so it should still be tested and should still work.

qaxi:
Use Python 2.7 is dangerous because there are not any security updates.

I agree with the second sentence, but I do not know how to solve it, because error I am correcting will make a stop for usage on Python 3.9 ... Do you have an idea how to change the push request to work on both - Python 2.7 and 3.9?

ntwrkguru:
> Do you have an idea how to change the push request to work on both - Python 2.7 and 3.9?

Probably need a conditional import statement based on the version...maybe like (untested, just off the cuff):

```python
import sys
if sys.version_info[0] == 2:
    import collections
else:
    import collections.abc
```

qaxi:
This pull request was siting here so long that it become obsolete ...
I have to watch notification more closely ... ;-)

This solved the problem ...
https://github.com/Juniper/py-junos-eznc/pull/1035/commits/74631fc97627bac8d7d40fb8ab4d382269f6b16f

``` python
try:
    from collections.abc import MutableMapping
except ImportError:
    from collections import MutableMapping
```

ntwrkguru:
That's one way to do it. :-)

Issue #1099, opened 2021-03-24
Error "AttributeError: 'NoneType' object has no attribute 'group'" when installing software on switches with multiple RE


wilamike:
Return error when calling pyEZ using Ansible module juniper.device.juniper_junos_software. It only happens for Juniper EX switch with more than 1 RE. The test unit is a Juniper EX4300 with 2 REs on OS 18.2R1.9. 

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: AttributeError: 'NoneType' object has no attribute 'group'
fatal: [junipersw]: FAILED! => changed=false
  module_stderr: |-
    Traceback (most recent call last):
      File "/home/mikng/.ansible/tmp/ansible-local-28385bnn35ozf/ansible-tmp-1616549333.4061427-28471-77421281602831/AnsiballZ_juniper_junos_software.py", line 102, in <module>
        _ansiballz_main()
      File "/home/mikng/.ansible/tmp/ansible-local-28385bnn35ozf/ansible-tmp-1616549333.4061427-28471-77421281602831/AnsiballZ_juniper_junos_software.py", line 94, in _ansiballz_main
        invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
      File "/home/mikng/.ansible/tmp/ansible-local-28385bnn35ozf/ansible-tmp-1616549333.4061427-28471-77421281602831/AnsiballZ_juniper_junos_software.py", line 40, in invoke_module
        runpy.run_module(mod_name='ansible_collections.juniper.device.plugins.modules.juniper_junos_software', init_globals=None, run_name='__main__', alter_sys=True)
      File "/usr/lib64/python3.6/runpy.py", line 205, in run_module
        return _run_module_code(code, init_globals, run_name, mod_spec)
      File "/usr/lib64/python3.6/runpy.py", line 96, in _run_module_code
        mod_name, mod_spec, pkg_name, script_name)
      File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
        exec(code, run_globals)
      File "/tmp/ansible_juniper.device.juniper_junos_software_payload_l35mb0t6/ansible_juniper.device.juniper_junos_software_payload.zip/ansible_collections/juniper/device/plugins/modules/juniper_junos_software.py", line 775, in <module>
[...]

PR #1101, opened 2021-04-02
Fix issue #1099

wilamike:
Fix issue #1099 . The logic is to skip values of self._RE_list which does not contain any digits.


rahkumar651991 (approved):
looks fine

PR #1110, opened 2021-04-26
fix xml iterator

bnavigator:
`xml.etree.ElementTree.getiterator()` was deprecated in Python 3.2 and removed in Python 3.9:

https://docs.python.org/3.8/library/xml.etree.elementtree.html#xml.etree.ElementTree.ElementTree.getiterator


rahkumar651991 (approved):
looks fine

PR #1116, opened 2021-06-02
Fix for issue #1093

dkasten79:
Added check for rsp variable in method diff to check for variable being a bool and then checking that the value is True before returning None which means no changes in the configuration for certain Junos releases.

### DIFF
--- a/lib/jnpr/junos/utils/config.py
+++ b/lib/jnpr/junos/utils/config.py
@@ -268,7 +268,7 @@ class Config(Util):
                 "diff shouldn't return boolean as a rpc-reply",
                 RuntimeWarning,
             )
-            return None
+            return None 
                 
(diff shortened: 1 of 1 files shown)


rahkumar651991:
@dkasten79 @dangoscomb - I discussed with my team. It is an error scenario to get bool in the rpc-reply and is present in some of the versions of Junos. Have mentioned the PR(internal to junos) in the comment above the fix. 

I have made few changes to this pull request for the same. This bug was also caught by one of the PyEZ user so we need to atleast inform the user about it. Displaying a warning but not returning an error and the preventing check will return None. 

Can one of you try this and let me know if it is fine or if you have any concerns ?

dangoscomb:
This seems sensible to me.

rahkumar651991 (approved):
looks ok.

PR #1129, opened 2021-08-19
Use cases

hacktivism-github:
Network Engineers with IOS/XR background used to deal with text files find their life a lot easier moving onto Junos world with such a solution.


rahkumar651991:
Hi @hacktivism-github - Since this is an example usage, juniper has another repository with multiple sample files. It is recommended to post it there. 
You can create a directory at this link and submit the pull request there -https://github.com/Juniper/junosautomation/tree/master/pyez

Closing this pull request here.

Issue #1135, opened 2021-10-28
Cluster ID of 16 fails to properly identify as node0 or node1 in current_re


blkangl:
https://github.com/Juniper/py-junos-eznc/blob/9b6a706b28d9be081c036931822ed0f92085b29c/lib/jnpr/junos/facts/current_re.py#L69

When we have a cluster ID of "16", the new code returns cluster_id_octet as "0"... but the ip of node1 is 130.16.0.1. Did something change in junos that breaks this new code? From what I see, this new code will always return cluster_id_octet as '0' when the id is 16 (or any multiple of 16)

By changing line 66 to:

                            cluster_id_octet = str(
                                (int(device.facts["srx_cluster_id"]) & 0x000F) << 4
                            )
                            if cluster_id_octet == "0":
                                cluster_id_octet = str(device.facts["srx_cluster_id"])
This allows my cluster with id "16" to function. I am not sure if this will work for all cases where the cluster ID binary ends in 0000.

chidanandpujar:
Verified the above fix ,with cluster_id 2 and cluster_id 16  ,for both the cluster ids , the above fix works fine and current_re  vlaues are populated according to cluster node details .

I will continue the testing for few more cluster id values and update the results .

from jnpr.junos import Device
from pprint import pprint

device = Device(host='xx.xx.xx.xx', user='xyz', password='zyx')
device.open()

pprint(device.facts["srx_cluster_id"])

cluster_id_octet = str(
     (int(device.facts["srx_cluster_id"]) & 0x000F) << 4
)
pprint(cluster_id_octet)
pprint(device.facts["current_re"])


~/junos-pyez# python get-facts.py 
'16'
'0'
['node0',
 'master',
 'fpc0',
 'node',
 'fwdd',
 'member',
 'pfem',
 're0',
 'fpc0.pic0']


~/junos-pyez# python get-facts.py 
'16'
'0'
['node1',
 'master',
 'fpc0',
 'node',
 'fwdd',
 'member',
 'pfem',
 're0',
 'fpc0.pic0']




with cluster id 2:

~/junos-pyez# python get-facts.py 
'2'
'32'
['node0',
 'master',
 'fpc0',
 'node',
 'fwdd',
 'member',
 'pfem',
 're0',
 'fpc0.pic0']


/junos-pyez# python get-facts.py 
'2'
'32'
['node1',
 'master',
 'fpc0',
 'node',
 'fwdd',
 'member',
 'pfem',
[...]

PR #1137, opened 2021-11-17
fixed issue #1135

chidanandpujar:
cluster_id_octet value is set to '0' when srx_cluster_id is 16 or multiples of 16 .
updated the get_facts , if cluster_id_octet is '0' ,  cluster_id_octet is assigned with srx_cluster_id  string .


dineshbaburam91 (approved):
Looks good

Issue #995, opened 2020-02-27
jnpr.junos.utils.start_shell option to choose shell (sh or csh)


cserranobr:
Hello All,

Good afternoon!

I've  already talked to Nitin Kumar about this issue.

I found out that StartShell only has csh, but the CLI command supports sh or csh  as the shell options (I always preferred to work with sh):
 
And the CLI command has both options:
 
root@c21-20> start shell ?
Possible completions:
  <[Enter]>            Execute this command
  csh                  Start C shell
  sh                   Start Bourne-style shell (really ash)
  user                 Start shell as another user
  |                    Pipe through a command
{master:0}
 
Thanks,
Carlos.

chidanandpujar:
Fix is merged https://github.com/Juniper/py-junos-eznc/pull/1173

PR #1147, opened 2022-01-03
Fix for issue #995

chidanandpujar:
Fix for issue  jnpr.junos.utils.start_shell option to choose shell (sh or csh) #995 
with this fix  StartShell supports the argument  shell_type for on-box  , we can pass csh or sh  shell types

### DIFF
--- a/lib/jnpr/junos/utils/start_shell.py
+++ b/lib/jnpr/junos/utils/start_shell.py
@@ -3,6 +3,8 @@ import re
 import datetime
 from jnpr.junos.utils.ssh_client import open_ssh_client
 import subprocess
+import six
+from threading import Thread
 
 _JUNOS_PROMPT = "> "
 _SHELL_PROMPT = "(%|#|\$)\s"
@@ -23,7 +25,7 @@ class StartShell(object):
 
     """
 
-    def __init__(self, nc, timeout=30):
+    def __init__(self, nc, timeout=30, shell_type='csh'):
         """
         Utility Constructor
 
@@ -37,6 +39,12 @@ class StartShell(object):
         self._client = None
         self._chan = None
         self.ON_JUNOS = self._nc.__class__.ON_JUNOS
+        self.shell_type = shell_type
+
+    def write_stdin(self, stdin, data):
+        while True:
+            stdin.write(six.b(data))
+            stdin.flush()
 
     def wait_for(self, this=_SHELL_PROMPT, timeout=0, sleep=0):
         """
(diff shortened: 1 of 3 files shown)


lib/jnpr/junos/utils/start_shell.py
```diff
@@ -100,18 +112,20 @@ def open(self):
...
+                bufsize=1,
             )
         else:
             self._client = open_ssh_client(dev=self._nc)
             self._chan = self._client.invoke_shell()
 
             got = self.wait_for(r"(%|>|#|\$)")
             if got[-1].endswith(_JUNOS_PROMPT):
-                self.send("start shell")
+                self.send("start shell self.shell_type")
```
dineshbaburam91, line 128: You need to concatenate two string

lib/jnpr/junos/utils/start_shell.py
```diff
@@ -100,18 +112,20 @@ def open(self):
         """
         if self.ON_JUNOS is True:
             self._chan = subprocess.Popen(
-                ["cli", "start", "shell"],
+                ["cli", "start", "shell", self.shell_type],
                 shell=False,
                 stdin=subprocess.PIPE,
                 stdout=subprocess.PIPE,
+                close_fds=1,
```
dineshbaburam91, line 119: Where did you create fd ?

lib/jnpr/junos/utils/start_shell.py
```diff
@@ -100,18 +112,20 @@ def open(self):
         """
         if self.ON_JUNOS is True:
             self._chan = subprocess.Popen(
-                ["cli", "start", "shell"],
+                ["cli", "start", "shell", self.shell_type],
```
dineshbaburam91, line 115: What is the behaviour if start shell is None

lib/jnpr/junos/utils/start_shell.py
chidanandpujar: Thanks for the review comments , 
Fixed the command 
self.send("start shell " + self.shell_type)

lib/jnpr/junos/utils/start_shell.py
chidanandpujar: removed the close_fds and bufsize
            self._chan = subprocess.Popen(
                ["cli", "start", "shell", self.shell_type],
                shell=False,
                stdin=subprocess.PIPE,
                stdout=subprocess.PIPE,
            )

lib/jnpr/junos/utils/start_shell.py
chidanandpujar: shell_type="csh"  is initialized with csh by default in init of StartShell,

dineshbaburam91:
Please update UT.

Issue #1184, opened 2022-06-23
utils.sw.install gives int() argument must be a string, a bytes-like object or a number, not 'NoneType'


Aaron-MJohn:
I am trying to upgrade the EX2300 with 48 POE+ switch using below code. I don't set any time out parameters in install function. Everything is using the default value.

```
with Device(host=hostname, user=junos_username, passwd=junos_password) as dev:
        print("Connecting to {}".format(hostname))
        if dev.connected:
            print("Device connected")
            sw = SW(dev)`
            ok = sw.install(
                package=firware_update_files,
                progress=my_progress,
                no_copy=True,
                validate=False,
                cleanfs=True,
            )
```
I get this exception 

```
192.168.1.2:request-package-checks-pending-install rpc is not supported on given device
192.168.1.2:installing software ... please be patient ...
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/jnpr/junos/device.py", line 227, in timeout
    self._conn.timeout = int(value)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/jnpr/junos/decorators.py", line 24, in wrapper
    dev.timeout = restore_timeout
  File "/usr/local/lib/python3.8/dist-packages/jnpr/junos/device.py", line 229, in timeout
    raise RuntimeError(
RuntimeError: could not convert timeout value of None to an integer

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/jnpr/junos/device.py", line 227, in timeout
    self._conn.timeout = int(value)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/..../devel/ros_ws/src/bg_sandbox/juno_config/scripts/juno_update.py", line 62, in <module>
    main()
  File "/home/..../devel/ros_ws/src/bg_sandbox/juno_config/scripts/juno_update.py", line 37, in main
    ok = sw.install(
  File "/usr/local/lib/python3.8/dist-packages/jnpr/junos/utils/sw.py", line 990, in install
    add_ok = self.pkgadd(
  File "/usr/local/lib/python3.8/dist-packages/jnpr/junos/utils/sw.py", line 240, in pkgadd
[...]

PR #1192, opened 2022-07-20
Fix for issue #1184

chidanandpujar:
Fix for timeout value getting set to None , changed it to default 30 seconds, when not set .


dineshbaburam91 (approved):
Looks good

PR #1199, opened 2022-08-17
fix reboot failing on other RE

apurvaraghu:
<request-reboot> tag is missing from Pyez. Appended the tag in order to reboot other RE successfully.


dineshbaburam91 (approved):
Looks good

Issue #1203, opened 2022-08-26
StartShell UnboundLocalError: local variable 'data' referenced before assignment 


ipmonk:
getting an exception when calling StartShell run() using PyEZ 2.6.4+. 
There is no data in the stdout buffer to read for several iterations of the loop...

```
#!/usr/bin/env python3
from jnpr.junos import Device
from jnpr.junos.utils.start_shell import StartShell

def try_run():
    dev = Device("foobar", user="foo", passwd="bar",port=22)
    with StartShell(dev) as ss:
        result = ss.run("hostname")
        print(result)
if __name__ == "__main__":
    try_run()
```

```
Traceback (most recent call last):
  File "repro.py", line 24, in <module>
    try_run()
  File "repro.py", line 21, in try_run
    result = ss.run("hostname")
  File "/Users/foobar/scripts/pyez_test_scripts/venv/lib/python3.6/site-packages/jnpr/junos/utils/start_shell.py", line 185, in run
    got = "".join(self.wait_for(this, timeout, sleep=sleep))
  File "/Users/foobar/scripts/pyez_test_scripts/venv/lib/python3.6/site-packages/jnpr/junos/utils/start_shell.py", line 82, in wait_for
    if isinstance(data, bytes):
UnboundLocalError: local variable 'data' referenced before assignment
```

Possible solution:

```
                data = b""       <<<< add this line
                rd, wr, err = select([chan], [], [], _SELECT_WAIT)
                if rd:
                    data = chan.recv(_RECVSZ)
```

sanjay75409:
I was getting same error then I downgraded to version 2.6.3

chidanandpujar:
Fix is under review https://github.com/Juniper/py-junos-eznc/pull/1207

chidanandpujar:
Hi,

 Could you please share the steps reproduce this issue .
issue is not reproducible with about script .

Thanks & Regards
Chidanand

ipmonk:
Hi, 

It reproduces in v2.6.4 no problem. It will not reproduce after the commit made on Sep 2nd for issue #1207. 

(venv-3.10.4) foo@foo-mbp venv-3.10.4 % python3 foo.py
Traceback (most recent call last):
  File "/Users/foo/venv-3.10.4/foo.py", line 11, in <module>
    try_run()
  File "/Users/foo/venv-3.10.4/foo.py", line 7, in try_run
    with StartShell(dev) as ss:
[...]

PR #1207, opened 2022-09-02
Fix for issue 1203

chidanandpujar:
Fix for issue #1203 
initialized the variable data .


lib/jnpr/junos/utils/start_shell.py
```diff
@@ -73,6 +74,7 @@ def wait_for(self, this=_SHELL_PROMPT, timeout=0, sleep=0):
             if self.ON_JUNOS is True:
                 data = chan.stdout.readline()
             else:
+                data = b""
```
dineshbaburam91, line 77: Update UT

Issue #1234, opened 2023-02-13
PyEZ connection fails when look_for_keys is default set to True and key_filenames is set to empty list


chidanandpujar:
PyEZ connection fails when look_for_keys is default set to True and key_filenames is set to empty list .
Could not add a 'pyez' channel due to: ConnectAuthError(xx.xx.xx.xx)

PR #1235, opened 2023-02-20
Fix for issue #1234

chidanandpujar:
option to set look_for_keys


lib/jnpr/junos/device.py
```diff
@@ -1215,6 +1215,10 @@ def __init__(self, *vargs, **kvargs):
             *OPTIONAL* parse XML with very deep trees and long text content.
             default is ``False``.
 
+        :param bool look_for_keys:
```
dineshbaburam91, line 1218: Please write UTP

dineshbaburam91 (approved):

Issue #1206, opened 2022-08-30
sw.install giving parse error when running on windows system


Aaron-MJohn:
Getting this error when running this code on windows system but works on Linux system.  
```
ok, msg = sw.install(
                        package=firmware_file,
                        progress=True,
                        no_copy=False,
                        validate=False,
                        cleanfs=True,
                    )
```
where firmware_file is file path to the firmware on the host system.

`fetch-secure: C:\Users\****\Desktop\****\network_switch_tool\ex2300-c\junos-arm-32-20.2R3-S3.6.tgz: parse error
ERROR: Cannot fetch file: C:\Users\****\Desktop\****\network_switch_tool\ex2300-c\junos-arm-32-20.2R3-S3.6.tgz`

Aaron-MJohn:
Thanks you. 
My current solution is scp the file and update it from locally from the switch.

chidanandpujar:
Fixed and merged via https://github.com/Juniper/py-junos-eznc/pull/1242

PR #1242, opened 2023-04-03
Fix for issue #1206

chidanandpujar:
Fix for issue https://github.com/Juniper/py-junos-eznc/issues/1206 ,
sw.install fails due to urlparase of C:\Users*\Desktop*\network_switch_tool\ex2300-c\junos-arm-32-20.2R3-S3.6.tgz , considers it as remote url since it contains C: , ignores the no_copy=False and the pkg file is not copied to the device and directly tries to download from the remote-url and fails with error .

Added check for remote url and if it is windows path C: , then safe_copy will be invoked to copy the pkg to the device

### DIFF
--- a/lib/jnpr/junos/utils/sw.py
+++ b/lib/jnpr/junos/utils/sw.py
@@ -920,9 +920,10 @@ class SW(Util):
         ):
             pkg_set = [package]
         if isinstance(pkg_set, (list, tuple)) and len(pkg_set) > 0:
+            remote_urls = ("ftp", "scp", "http", "https", "tftp")
             for pkg in pkg_set:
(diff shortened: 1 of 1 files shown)


lib/jnpr/junos/utils/sw.py
```diff
@@ -920,9 +920,10 @@ def _progress(report):
         ):
             pkg_set = [package]
         if isinstance(pkg_set, (list, tuple)) and len(pkg_set) > 0:
+            remote_urls = ["ftp", "scp", "http", "https", "tftp"]
```
dineshbaburam91, line 923: Add SFTP as well

dineshbaburam91 (approved):

Issue #1280, opened 2023-11-04
Unable to install PyEZ in MACOS ventura 13.6.1


dhotresumit:
I am getting an error while installing PyEZ module on Mac OS 13.6.1 and Python 3.12.0
>>> from jnpr.junos import Device
Traceback (most recent call last):
File “<stdin>”, line 1, in <module>
ModuleNotFoundError:

Please help me to get started.

dhotresumit:
Hi Team, any update on this?

dineshbaburam91:
PyEZ support python 3.10 and above version

Issue #1276, opened 2023-10-22
Support py3.12


chenrui333:
There is some build issue with 2.6.8 release build against py3.12

```
Using pip 23.2.1 from /opt/homebrew/lib/python3.12/site-packages/pip (python 3.12)
Processing /private/tmp/ansible--junos-eznc-20231020-28117-bbknce/junos-eznc-2.6.8
  Preparing metadata (setup.py): started
  Running command python setup.py egg_info
  /private/tmp/ansible--junos-eznc-20231020-28117-bbknce/junos-eznc-2.6.8/versioneer.py:432: SyntaxWarning: invalid escape sequence '\s'
    ] = '''
  Traceback (most recent call last):
    File "<string>", line 2, in <module>
    File "<pip-setuptools-caller>", line 34, in <module>
    File "/private/tmp/ansible--junos-eznc-20231020-28117-bbknce/junos-eznc-2.6.8/setup.py", line 11, in <module>
      version=versioneer.get_version(),
              ^^^^^^^^^^^^^^^^^^^^^^^^
    File "/private/tmp/ansible--junos-eznc-20231020-28117-bbknce/junos-eznc-2.6.8/versioneer.py", line 1524, in get_version
      return get_versions()["version"]
             ^^^^^^^^^^^^^^
    File "/private/tmp/ansible--junos-eznc-20231020-28117-bbknce/junos-eznc-2.6.8/versioneer.py", line 1451, in get_versions
      cfg = get_config_from_root(root)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/private/tmp/ansible--junos-eznc-20231020-28117-bbknce/junos-eznc-2.6.8/versioneer.py", line 346, in get_config_from_root
      parser = configparser.SafeConfigParser()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  AttributeError: module 'configparser' has no attribute 'SafeConfigParser'. Did you mean: 'RawConfigParser'?
  error: subprocess-exited-with-error
```

full build log, https://github.com/Homebrew/homebrew-core/actions/runs/6582900775/job/17885208567
relates to https://github.com/Homebrew/homebrew-core/pull/151733

ljarasius:
Issue is related to the NETCONF library underpinning PyEz. We have been getting around this issue by continuing to use Python 3.11 until ncclient has an update, which was completed in the last few days.

https://github.com/ncclient/ncclient/issues/572

[...]

Issue #1253, opened 2023-07-26
python 3.10 support for PyEZ


spithavath:
Testing Python 3.10 support for PyEZ

chidanandpujar:
with nose2 we are seeing multiple errors , we are looking into it .

```
======================================================================
ERROR: test_sw_info_bsys (tests.unit.facts.test_get_software_information.TestGetSoftwareInformation)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/pyez_release_py310_latest/venv/lib/python3.10/site-packages/mock/mock.py", line 1452, in patched
    return func(*newargs, **newkeywargs)
  File "/root/pyez_release_py310_latest/py-junos-eznc/tests/unit/facts/test_get_software_information.py", line 171, in test_sw_info_bsys
    self.assertEqual(self.dev.facts["hostname"], "bsys")
  File "/root/pyez_release_py310_latest/venv/lib/python3.10/site-packages/junos_eznc-2.6.7+27.gba799a81-py3.10.egg/jnpr/junos/factcache.py", line 119, in __getitem__
    raise RuntimeError(
RuntimeError: The jnpr.junos.facts.get_software_information module returned the model fact, but does not list model as a provided fact. Please report this error.

======================================================================
ERROR: test_sw_info_dual (tests.unit.facts.test_get_software_information.TestGetSoftwareInformation)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/pyez_release_py310_latest/venv/lib/python3.10/site-packages/mock/mock.py", line 1452, in patched
    return func(*newargs, **newkeywargs)
  File "/root/pyez_release_py310_latest/py-junos-eznc/tests/unit/facts/test_get_software_information.py", line 104, in test_sw_info_dual
    self.assertEqual(self.dev.facts["junos_info"]["re0"]["text"], "15.1F5.15-C1.12")
  File "/root/pyez_release_py310_latest/venv/lib/python3.10/site-packages/junos_eznc-2.6.7+27.gba799a81-py3.10.egg/jnpr/junos/factcache.py", line 119, in __getitem__
    raise RuntimeError(
RuntimeError: The jnpr.junos.facts.get_software_information module returned the model fact, but does not list model as a provided fact. Please report this error.

======================================================================
ERROR: test_sw_info_dual_other_re_off (tests.unit.facts.test_get_software_information.TestGetSoftwareInformation)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

Issue #1252, opened 2023-07-18
inspect.getargspec is deprecated and removed in Python 3.11


daemonkeeper:
Python 3.11 removed the inspect.getargspec function from the inspect module. This function is used in the SCP module here: https://github.com/Juniper/py-junos-eznc/blob/a106028765d791a515229d422ddfbdfed5c08e0c/lib/jnpr/junos/utils/scp.py#L46

Please replace the invokation by inspect.getfullargspec which acts as a drop-in replacement as per https://docs.python.org/3/library/inspect.html#inspect.getfullargspec.

Since getfullargspec is available since the early days of Python 3.x (and definitely available in Python 3.5 which you name as minimum version) you probably don't need to bump your python dependencies unless you plan to support Python 2. 

edit: the same probably applies to https://github.com/Juniper/py-junos-eznc/blob/a106028765d791a515229d422ddfbdfed5c08e0c/lib/jnpr/junos/device.py#L49 

However there, it looks, you have remainders of Python 2 compatibility code in place which can probably stay around.

ludal35:
Hello - same problem occurs with while copying files:

```
File "/usr/local/lib/python3.11/dist-packages/jnpr/junos/utils/sw.py", line 936, in install
    copy_ok = self.safe_copy(
              ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/jnpr/junos/utils/sw.py", line 671, in safe_copy
    self.put(package, remote_path=remote_path, progress=progress)
  File "/usr/local/lib/python3.11/dist-packages/jnpr/junos/utils/sw.py", line 203, in put
    with SCP(self._dev, progress=progress) as scp:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/jnpr/junos/utils/scp.py", line 46, in __init__
    spec = inspect.getargspec(self._user_progress)
           ^^^^^^^^^^^^^^^^^^
AttributeError: module 'inspect' has no attribute 'getargspec'. Did you mean: 'getargs'?
```

See https://github.com/Juniper/py-junos-eznc/blob/53054ea685951f1726c521305a4eb87e21839f91/lib/jnpr/junos/utils/scp.py#L46

dineshbaburam91:
pyEZ support python3.10 and above

PR #1281, opened 2023-11-06
python3.12 support

chidanandpujar:
Python3.12 support
#1253
#1252 
#1280
#1276

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 lxml>=3.2.4
 # ncclient version 0.6.10 has issues with PyEZ(junos-eznc) and needs to be avoided
-ncclient==0.6.13
+ncclient==0.6.15
 paramiko>=1.15.2
 scp>=0.7.0
 jinja2>=2.7.1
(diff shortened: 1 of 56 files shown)


requirements.txt
```diff
@@ -1,6 +1,6 @@
 lxml>=3.2.4
 # ncclient version 0.6.10 has issues with PyEZ(junos-eznc) and needs to be avoided
-ncclient==0.6.13
+ncclient==0.6.15
```
ktbyers, line 3: Really should make this a `ncclient>=0.6.15` (to try avoid pinning to a specific version of ncclient).

chidanandpujar: Hi @ktbyers 
    Thanks vey much for the review comments .
   I will update the fix .

Thanks

ktbyers:
We should update this (so newer versions of Python have some testing in the CI/CD system):

https://github.com/Juniper/py-junos-eznc/blob/master/.github/workflows/pylint.yml#L17

@chidanandpujar Do you want to add that into this PR?

chidanandpujar:
> We should update this (so newer versions of Python have some testing in the CI/CD system):
> 
> https://github.com/Juniper/py-junos-eznc/blob/master/.github/workflows/pylint.yml#L17
> 
> @chidanandpujar Do you want to add that into this PR?

Hi @ktbyers 
    Thanks for the review comments .
    We will update the github workflow with never versions on python.

Thanks

PR #1284, opened 2023-11-23
feat: Enable using only the SSH agent if asked.

gaima8:
Matching PR for https://github.com/Juniper/ansible-junos-stdlib/pull/634

PR #1285, opened 2023-11-25
Do not disable SSH agent

vincentbernat:
The current logic around the SSH agent is flawed and tentatives to fix it to cover more cases only leads to more complexity.

As seen in #648, to this day, nobody really knows why we disable the agent. In the meantime, many valid use cases just don't work because of disabling the agent, including the good practice to use encrypted private key files. I think it is time to bite the bullet and just leave the agent always on: "ssh" from OpenSSH does not have an option to disable the agent and nobody complains about this.

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -1354,15 +1354,6 @@ class Device(_Connection):
         try:
             ts_start = datetime.datetime.now()
 
-            # we want to enable the ssh-agent if-and-only-if we are
-            # not given a password or an ssh key file.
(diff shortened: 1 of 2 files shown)


gaima8:
Can confirm this works.
Better than my patches https://github.com/Juniper/py-junos-eznc/pull/1284 and https://github.com/Juniper/ansible-junos-stdlib/pull/634 as this only requires the one change here.

chidanandpujar:
merged https://github.com/Juniper/py-junos-eznc/pull/1388

PR #1290, opened 2024-01-11
add 'member' to reboot RPC only when it's not None

MicLennartz:
The current logic adds the 'member' attribute to the request-reboot RPC, even if the member attribute is 'None' (default) while the target device is running on a single-node virtual chassis.

This causes the RPC call to fail.

-> parentheses were missing to reflect the right logic

### DIFF
--- a/lib/jnpr/junos/utils/sw.py
+++ b/lib/jnpr/junos/utils/sw.py
@@ -1131,10 +1131,8 @@ class SW(Util):
             elif self._mixed_VC is True:
                 cmd.append(E("all-members"))
         elif (
-            (self._multi_VC_nsync is True
-            or self._multi_VC is True)
-            and member_id is not None
-        ):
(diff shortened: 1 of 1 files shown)


lib/jnpr/junos/utils/sw.py
```diff
@@ -1131,8 +1131,8 @@ def _system_operation(
             elif self._mixed_VC is True:
                 cmd.append(E("all-members"))
         elif (
-            self._multi_VC_nsync is True
-            or self._multi_VC is True
+            (self._multi_VC_nsync is True
```
dineshbaburam91, line 1134: Kindly fix the format for your fix by running "black" tools.  Workflow is failing with your fix

MicLennartz: Ok, format is now updated according to 'black'.

dineshbaburam91 (approved):

PR #1294, opened 2024-01-18
implement shutdown for vmhost devices

apurvaraghu:
Devices with VMHOST only had reboot action supported and would throw an error if shutdown/poweroff was performed.  The fix was to implement action:'shutdown' in the poweroff function with a check for vmhost=True and then call the respective rpc for shutting down vmhost devices on PyEz side. Also modify the function call for poweroff on the anisble side to support the same.


dineshbaburam91 (commented):
Looks good

dineshbaburam91 (approved):

apurvaraghu:
fixes #587

PR #1303, opened 2024-04-13
Fixed unit test test_textfsm_table_mutli_key failing because of invalid key error

Aaron-MJohn:
The template has "MAC_ADDRESS" as key but the reference field has only MAC which cause a invalid key and fails the unit test


dineshbaburam91:
This changes is not required

PR #1307, opened 2024-04-17
Don't require six

MeggyCal:
As this library is python3-only, it shouldn't need six.

If there are any mistakes in the patch, please let me know. It was partially done by far-from-perfect sed script.

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -1,6 +1,5 @@
 # stdlib
 import os
-import six
 import types
 import platform
 import warnings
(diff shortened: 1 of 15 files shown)


a-detiste:
Hi, I agree

a-detiste:
At the leatest, six is not use for literals ( six.b(literal) ) since Python 2.6 / 3.1 ...

dineshbaburam91:
This will be taken care of when PyEZ supports python 3.13 support. So, I'm closing this pull request.

PR #1311, opened 2024-05-03
Updated mock to built-in unittest.mock

dineshbaburam91:
Removed a `mock` dependency in the PyEZ module.
`mock` was a third-party package for Python 2.7 first, then became part of the standard library in Python 3, available as unittest.mock.  So, updated unit testcases to use unittest.mock module.


chidanandpujar (approved):
Looks good

Issue #1320, opened 2024-08-26
SSH connection hard-coded to NOT verify host keys


ed-ud:
Why is there a hard-coded "hostkey_verify=False" here, effectively disabling SSH host key verification...???

https://github.com/Juniper/py-junos-eznc/blame/a64698b280fb717730dddf75a2d4d8cacb2e4775/lib/jnpr/junos/device.py#L1377

chidanandpujar:
Hi @ed-ud,
    I have added the boolean parameter to read the hostkey_value  with True or False.

please find the test log with fix 

```
1) hostkey_verify True,
~/pyez_release_272_py39/py-junos-eznc# cat get_facts.py
from jnpr.junos import Device
from pprint import pprint

dev = Device(host='x.x.x.x', user='xyz', password='xyz', port=22, hostkey_verify=True)
dev.open()
pprint (dev.facts)

/pyez_release_272_py39/py-junos-eznc# python get_facts.py 
/root/pyez_release_272_py39/venv/lib/python3.9/site-packages/paramiko/pkey.py:100: CryptographyDeprecationWarning: TripleDES has been moved to cryptography.hazmat.decrepit.ciphers.algorithms.TripleDES and will be removed from this module in 48.0.0.
  "cipher": algorithms.TripleDES,
/root/pyez_release_272_py39/venv/lib/python3.9/site-packages/paramiko/transport.py:271: CryptographyDeprecationWarning: TripleDES has been moved to cryptography.hazmat.decrepit.ciphers.algorithms.TripleDES and will be removed from this module in 48.0.0.
  "class": algorithms.TripleDES,
Traceback (most recent call last):
  File "/root/pyez_release_272_py39/venv/lib/python3.9/site-packages/junos_eznc-2.7.1+9.ga64698b2.dirty-py3.9.egg/jnpr/junos/device.py", line 1388, in open
    self._conn = netconf_ssh.connect(
  File "/root/pyez_release_272_py39/venv/lib/python3.9/site-packages/ncclient-0.6.15-py3.9.egg/ncclient/manager.py", line 187, in connect
  File "/root/pyez_release_272_py39/venv/lib/python3.9/site-packages/ncclient-0.6.15-py3.9.egg/ncclient/manager.py", line 139, in connect_ssh
  File "/root/pyez_release_272_py39/venv/lib/python3.9/site-packages/ncclient-0.6.15-py3.9.egg/ncclient/transport/ssh.py", line 349, in connect
[...]

PR #1321, opened 2024-09-03
Option to enable SSH known host key verification

chidanandpujar:
Fix for #1320 
Provided boolean parameter hostkey_verify (False/True) to disable/enable SSH known hosts key verification during ssh connection.  By default it is disabled with value hostkey_verify=False.

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -1216,6 +1216,10 @@ class Device(_Connection):
             *OPTIONAL* To disable public key authentication.
             default is ``None``.
 
+        :param bool hostkey_verify:
+            *OPTIONAL* To enable ssh_known hostkey verify
+            default is ``False``.
+
         """
 
         # ----------------------------------------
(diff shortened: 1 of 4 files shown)


lib/jnpr/junos/device.py
```diff
@@ -1367,14 +1372,22 @@ def open(self, *vargs, **kvargs):
...
                 look_for_keys = self._look_for_keys
 
+            # option to enable ssh_known hosts key verification
+            # using hostkey_verify=True
+            # Default is disabled with hostkey_verify=False
+            if self._hostkey_verify is None:
+                hostkey_verify = False
+            else:
+                hostkey_verify = self._hostkey_verify
+
```
dineshbaburam91, line 1382: Add UTP for this usecases

chidanandpujar: UT test cases are added

dineshbaburam91 (approved):

Issue #1322, opened 2024-09-06
sw.install "validate" parameter is not passed on to ISSU and NSSU installation.


chidanandpujar:
validate option with True/False to enable/disable for Validating configuration against package during NSSU upgrade is not working.

```
ok, msg = sw.install(package=pkg,
                     validate=False,
                     remote_path='/var/tmp',
                     progress=myprogress,
                     no_copy=True,
                     force_host=False,
                     reboot=False,
                     issu=False,
                     nssu = True,
                     cleanfs=False,
                     vmhost=False,
                     force_copy=False,
                     timeout=7500)
```
RPC invoked for NSSU is not having the option no-validate when validate=False.
```
<request-package-nonstop-upgrade><package-name>/var/tmp/junos-arm-32-24.1R1.5.tgz</package-name></request-package-nonstop-upgrade>
```

chidanandpujar:
fixed via https://github.com/Juniper/py-junos-eznc/pull/1334, hence closing the issue

PR #1323, opened 2024-09-06
Fix for issue #1322 to pass no_validate

chidanandpujar:
sw.install parameter validate=False is not passed on to the NSSU and ISSU upgrade .
Added code to set no_validate option when validate=False

### DIFF
--- a/lib/jnpr/junos/utils/sw.py
+++ b/lib/jnpr/junos/utils/sw.py
@@ -984,7 +984,7 @@ class SW(Util):
                     kwargs.update({"no_validate": True})
 
             if issu is True:
-                if validate is False: # To Check validation is False
+                if validate is False:  # To Check validation is False
                     # Need to pass the no_validate option via kwargs
                     kwargs.update({"no_validate": True})
                 _progress("ISSU: installing software ... please be patient ...")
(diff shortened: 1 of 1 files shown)


lib/jnpr/junos/utils/sw.py
```diff
@@ -984,11 +984,17 @@ def _progress(report):
                     kwargs.update({"no_validate": True})
 
             if issu is True:
+                if validate is False:
```
dineshbaburam91, line 987: Why are you explicitly setting the validate option as True always?

Are there any reference document , validation always be "True" in case of ISSU and NSSU

chidanandpujar: when we set the option validate=False, we are not passing on this option to request system software add/request system software nonstop-upgrade/request system software add.

no-validate	
(Optional) When loading a software package or bundle with a different release, suppress the default behavior of the validate option.

validate	
(Optional) Validate the software package or bundle against the current configuration as a prerequisite to adding the software package or bundle. This is the default behavior when the software package or bundle being added is a different release.

no-validate	
(Optional) Starting in Junos OS Release 20.4R2, skip the validation step of the unified ISSU process.

chidanandpujar:
Issue is taken care in the pull request https://github.com/Juniper/py-junos-eznc/pull/1334

PR #1352, opened 2024-12-16
Fix PY3.13 support issues; drop PY3.8 support

ktbyers:
PY3.8 is EOL.

Used vendored version of telnetlib inside of Netmiko; alternative is to completely drop telnet support (of find some alternatively maintained telnet library).


ktbyers:
Another option is to just gracefully catch the `import telnetlib` failure such that it doesn't cause the `Device` import to fail.

Basically, create a deprecation warning and gracefully catch the `import telnetlib` failure. In this way, people using PY3.9 - PY3.12 could keep using telnetlib and PY3.13 would work for its main uses. telnet function would cease working in PY3.13 as it would be deprecated by PyEZ.

a-detiste:
https://github.com/youknowone/python-deadlib provides telnetlib

dineshbaburam91:
@ktbyers  The “telnetlib” module is no longer supported in Python 3.13, so we must incorporate telnetlib3 instead. Supporting telnetlib3 in PyEZ is an enhancement, and I am currently analyzing the code.

Also in PyPI it's already updated saying it supports python3.12.
https://pypi.org/project/junos-eznc/

Also, we are working on updating the same in the juniper document.

ktbyers:
@dineshbaburam91 Multiple places have back-ported telnetlib (including Netmiko referenced in this PR). There is no need to support an asyncio version of telnet; that is a much bigger and unnecessary change.

At a minimum, don't make PyEZ fail when the telnetlib import fails (as is the current state)--i.e. just catch the exception on the `import telnetlib` fail and create a flag that says telnetlib not supported. The vast majority of users don't care about telnet support.

You could also just vendor the previous telnetlib.py code inside of PyEZ. It is a single Python module; this is a much smaller change, than somehow having an asyncio version of telnet inside of PyEZ. This is what I have done in Netmiko.

Yes I recognize you support PY3.12, but PY3.13 is a released version of Python and users are using it (and PyEZ breaks when people try to use it). PyEZ is also breaking other libraries (for example, NAPALM cannot support PY3.13 due to PyEZ).

PR #1356, opened 2025-01-21
Logging for paramiko and connection failures

chidanandpujar:
Enabled logging with paramiko.uit.log_file and logging.exception.

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -56,6 +56,7 @@ logger = logging.getLogger("jnpr.junos.device")
 logging.thread = None
 paramiko.util.log_to_file("paramiko.log")
 
+
 class _MyTemplateLoader(jinja2.BaseLoader):
     """
     Create a jinja2 template loader class that can be used to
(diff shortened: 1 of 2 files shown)


lib/jnpr/junos/device.py
```diff
@@ -52,6 +52,10 @@
 
 logger = logging.getLogger("jnpr.junos.device")
 
+# Setup paramiko logging
```
dineshbaburam91, line 55: You can get depend library logs using the snippet below. Adding separate logging files for third-party libraries is not acceptable.
```
import logging
 logging.basicConfig(level=logging.DEBUG)
```

dineshbaburam91 (commented):
You can use the snippet to get a depend library  tracing log. 

Adding third party logging infra code is not acceptable.
```
import logging
logging.basic config(level=logging.DEBUG)
```

chidanandpujar:
functionality can be achieved by adding the following lines to the PyEZ script.
import logging
 logging.basicConfig(level=logging.DEBUG)

Issue #1385, opened 2025-09-24
Allow ssh-agent-based certificate authentication


bcollet:
[ncclient v0.7.0](https://github.com/ncclient/ncclient/releases/tag/v0.7.0) added support to allow supplementing a public key found in an SSH agent with certificate data, if present.

This is useful for certificate authentication when the key material is only accessible through an SSH agent (for PKCS11-backed certificates for instance).

However, the current logic in PyEZ prevents using an ssh-agent while also loading a key or a certificate.

Other similar issues with this logic have been raised in the past, and forcing the ssh-agent to be disabled in certain conditions seems to cause more issues than it solves, if any:
- #1285 
- #1284
- #1353

chidanandpujar:
merged https://github.com/Juniper/py-junos-eznc/pull/1388

Issue #1353, opened 2024-12-18
ConnectAuthError when trying to connect to SRX with SSH private key with passphrase and SSH-agent


Raimond56:
Hello,

I need some help to troubleshoot why connecting with SSH key with passphrase is not working with SSH agent. 
If I understand correctly if I have setup access to SRX with SSH keys and using SSH agent then junos-eznc should be able to read that information and connect to device. 

Connection to SRX 4200 with version Junos: 21.4R3-S4.9
From ubuntu 22.04 with python 3.10 and junos-eznc 2.7.1

This is what I have in my .ssh/config file:
```
host 10.35.255.1
        IdentityFile ~/.ssh/privatekey
host *
        KexAlgorithms=+"diffie-hellman-group1-sha1,diffie-hellman-group14-sha1"
        PubkeyAcceptedAlgorithms=+"ssh-rsa"
        HostKeyAlgorithms=+"ssh-rsa"
        Ciphers=+"aes128-cbc,3des-cbc,aes192-cbc,aes256-cbc"
        ServerAliveInterval=300
        ServerAliveCountMax=2
AddKeysToAgent yes
```
If I connect from command line it doesn't ask password and connects

```
rilves@raimondadmin:~$ ssh 10.35.255.1 -p 830 -s netconf
<!-- No zombies were killed during the creation of this user interface -->
<!-- user rilves, class j-super-user-local -->
<hello xmlns="urn:ietf:params:xml:ns:netconf:base:1.0">
```

If I try with the following python code

```
from pprint import pprint
from jnpr.junos import Device

with Device(host='10.35.255.1', user='rilves') as dev:
    pprint( dev.facts )
```

       
I get following error:

```
rilves@raimondadmin:~$ python3 test.py
Traceback (most recent call last):
  File "/home/rilves/.local/lib/python3.10/site-packages/jnpr/junos/device.py", line 1371, in open
    self._conn = netconf_ssh.connect(
  File "/home/rilves/.local/lib/python3.10/site-packages/ncclient/manager.py", line 187, in connect
    return connect_ssh(*args, **kwds)
  File "/home/rilves/.local/lib/python3.10/site-packages/ncclient/manager.py", line 139, in connect_ssh
    session.connect(*args, **kwds)
  File "/home/rilves/.local/lib/python3.10/site-packages/ncclient/transport/ssh.py", line 359, in connect
    self._auth(username, password, key_filenames, allow_agent, look_for_keys)
[...]

PR #1388, opened 2026-01-06
allow_agent param support

chidanandpujar:
Support for allow_agent parameter

https://github.com/Juniper/py-junos-eznc/issues/1385
https://github.com/Juniper/py-junos-eznc/pull/1285
https://github.com/Juniper/py-junos-eznc/issues/1353

### DIFF
--- a/lib/jnpr/junos/device.py
+++ b/lib/jnpr/junos/device.py
@@ -1217,6 +1217,14 @@ class Device(_Connection):
             *OPTIONAL* To disable public key authentication.
             default is ``None``.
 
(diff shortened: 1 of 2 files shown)


vincentbernat:
In almost 10 years, nobody demonstrated a need to disable the agent. This is a setting not available in the OpenSSH client. While the ability to make this configurable is better than nothing, this setting may not be exposed in dependent libraries or applications, making secure usage of PyEZ a pain (password-protected SSH keys or SSH keys on Yubikey for example). #1285 is a simpler fix.

Also see #648.

